PR #1, opened 2020-10-08
Add first Dockerfile implementation

buehner:
Based on https://github.com/geoserver/geoserver/wiki/GSIP-192 this is a first step towards an official GeoServer docker image.

This is **work in progress**, but some features are already implemented:

* Extensions specified in the `STABLE_EXTENSIONS` environment variable will be downloaded and installed on container startup (before tomcat starts)
* Marlin renderer is used
* Recommended JVM production configuration

You can build with `docker build -t geoserver:test .`

You can run with `docker run -it -e STABLE_EXTENSIONS='wps,csw' -p 8080:8080 geoserver:test` (and pass a list of stable extensions).

More information is documented in the README.md.

### TODOs:

* [x] CORS support
* [ ] configuration of JNDI connections in the tomcat/custom tomcat configuration in general
* [ ] default data for gs datadir?
* [x] possibility to add custom java dependencies or extensions to WEB-INF/lib
  * [x] in this context: mount `EXTENSION_DOWNLOAD_DIR` and implement checks to cache extensions that have already been downloaded before?
* [ ] log4j properties
* [ ] possibility to add custom fonts
* [ ] backwards compatability to which version (maybe 2.15.x ?)
* [x] docker-compose demo environment

**Any feedback is appreciated!**


jodygarnett:
No need for special treatment for oracle jdbc driver anymore, oracle changed their policy and we have been able to depend on it as a maven dependency and redistribute now.

Dockerfile
```diff
@@ -0,0 +1,57 @@
+ARG TOMCAT_VERSION=9.0.38
+ARG JAVA_VERSION=8
```
mprins, line 2: I would stick to Java 11 which has Marlin built-in to save some hassle

README.md
```diff
@@ -1,2 +1,51 @@
...
+## TODOs
+
+* CORS
+* configuration of JNDI connections in the tomcat/custom tomcat configuration in general
+* default data for gs datadir?
+* possibility to add custom java dependencies (for example the oracle jdbc driver)
+  * in this context: mount `EXTENSION_DOWNLOAD_DIR` and implement checks to cache extensions that have already been downloaded before?
+* log4j properties
+* add possibility to add custom fonts
+* starting from which version we want to provide geoserver images (maybe 2.15.x?)/backwards compatability
```
mprins, line 50: supported versions are 17.x and up

scripts/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+export GEOSERVER_OPTS="-Djava.awt.headless=true -server \
+       -Dfile.encoding=UTF8 \
+       -Djavax.servlet.request.encoding=UTF-8 \
+       -Djavax.servlet.response.encoding=UTF-8 \
+       -Xms${INITIAL_MEMORY} -Xmx${MAXIMUM_MEMORY} \
+       -XX:SoftRefLRUPolicyMSPerMB=36000 -XX:+UseG1GC \
+       -XX:MaxGCPauseMillis=200 -XX:ParallelGCThreads=20 -XX:ConcGCThreads=5 \
+       -Xbootclasspath/a:${CATALINA_HOME}/lib/marlin-${MARLIN_VERSION}-Unsafe.jar \
+       -Dsun.java2d.renderer=org.marlin.pisces.MarlinRenderingEngine \
+       -Dorg.geotools.coverage.jaiext.enabled=${JAIEXT_ENABLED}"
```
mprins, line 19: some of these options may not be valid for Java 11 (don't know which off the top of my head)

README.md
```diff
@@ -1,2 +1,51 @@
...
+* `STABLE_EXTENSIONS` as comma separated list, will be downloaded and installed on startup (default: "")
+
+## TODOs
+
+* CORS
+* configuration of JNDI connections in the tomcat/custom tomcat configuration in general
+* default data for gs datadir?
+* possibility to add custom java dependencies (for example the oracle jdbc driver)
+  * in this context: mount `EXTENSION_DOWNLOAD_DIR` and implement checks to cache extensions that have already been downloaded before?
+* log4j properties
```
mprins, line 48: please set up a logging profile that does not spam standard out (`stdout`) like most of the defaults do

Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    STABLE_PLUGIN_URL=${STABLE_PLUGIN_URL} \
+    INITIAL_MEMORY="2G" \
+    MAXIMUM_MEMORY="4G" \
+    JAIEXT_ENABLED=true \
+    STABLE_EXTENSIONS='' \
+#    COMMUNITY_EXTENSIONS='' \
+    DEBIAN_FRONTEND=noninteractive \
+    EXTENSION_DOWNLOAD_DIR=/opt/geoserver_extension_downloads \
+    GEOSERVER_DATA_DIR=/opt/geoserver_data \
+    GEOWEBCACHE_CACHE_DIR=/opt/geowebcache_data
```
mprins, line 28: you may want to expose these as a volume so the can be backed up and migrated

README.md
jodygarnett: So you are recommending PRODUCTION.properties to start out with?

Dockerfile
jodygarnett: The GEOSERVER_DATA_DIR is handy for outside access, easy way to stage data for publication.

scripts/entrypoint.sh
jodygarnett: I tend to run with Java 8 and settings very similar to what you have here.

The default for memory is 1/4 system memory, not sure what the docker image ends up?

The `UseG1GC` is the default for Java 11, also marlin is baked in if using Java 11.

The number of threads to devote to garbage collection is is a tradeoff based on the number of threads available.

README.md
mprins: No, because `PRODUCTION.properties` still logs to `stdout` spamming the catalina logs / catalina.out /console with all it thinks we should know about: 

https://github.com/geoserver/geoserver/blob/d09be5b2ebfe0593418a9ce04668be0b17ec3d61/src/main/src/main/java/PRODUCTION_LOGGING.properties#L18

I'd prefer something like `PRODUCTION.properties` with `stdout` removed, so the catalina logs will just tell me things about the servlet engine and not geoserver (because that's what the geoserver log is for)

scripts/entrypoint.sh
mprins: > not sure what the docker image ends up

it depends if the container is created with resource limits like [`--memory`](https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory) if the container tries to allocate more than allowed it will be killed, but users can specify those using env variables 

https://github.com/geoserver/docker/blob/eee16c94e4a4fafdcb8b85b08f1583cb0efa6dbd/Dockerfile#L20-L21

https://github.com/geoserver/docker/blob/eee16c94e4a4fafdcb8b85b08f1583cb0efa6dbd/scripts/entrypoint.sh#L14

README.md
randomorder: I agree
Maybe we propose to change profile? I don't think it makes sense for production setups

scripts/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+## Inspired by https://github.com/kartoza/docker-geoserver
+set -e
+
+## install GeoServer extensions before starting the tomcat
+echo "Starting installation of extensions"
+/scripts/install-extensions.sh
```
randomorder, line 7: I would not install extensions at run time. For several reasons:
- does not work for community plugins
- it would not work in an environment without internet access
- extensions are downloaded every time the container is started. That's a lot of downloads
- should be a temporary issue with sourceforge geoserver will start without the necessary plugins and not function properly (even fail to start?)

I think GeoServer war bundling should just not happen here but instead should be handled separately and documented in the README. Users still have several options available to customize the application like bind mount a locally available war or override the command / enrtypoint to actually download extensions at startup if they want to or actually change the dockerfile and [ADD](https://docs.docker.com/engine/reference/builder/#add) or [COPY](https://docs.docker.com/engine/reference/builder/#copy) their war file directly into the docker image

beginor:
how about this geoserver docker image repo [beginor/docker-geoserver](https://github.com/beginor/docker-geoserver) and I have already pushed it on docker hub https://hub.docker.com/r/beginor/geoserver , you guys can just pickup it.

Dockerfile
buehner: I switched to OpenJDK 11 in a03ee5a

README.md
buehner: 17.x for technical reasons or because it was the latest version in Oct?

scripts/entrypoint.sh
buehner: I switched to OpenJDK 11 in a03ee5a which also adjusts the lines above

Dockerfile
buehner: Yes. These directories should be mounted. I'll soon add a docker-compose demo for this

scripts/entrypoint.sh
buehner: Yes. I agree that downloading and installing extensions at run time should not be the default, but i'd really like to keep this as a configurable feature (disabled by default) for quick setups.
 
I generally agree (regarding war bundling) that it should also be possible to build offline by mounting all required resources (like customized war files).

I'll work on this sooner or later and add some changes.

buehner:
> how about this geoserver docker image repo [beginor/docker-geoserver](https://github.com/beginor/docker-geoserver) and I have already pushed it on docker hub https://hub.docker.com/r/beginor/geoserver , you guys can just pickup it.

thx for your comment. if you miss any features, feel free to discuss and open pull requests

scripts/entrypoint.sh
buehner: I added a flag to control whether extensions should be downloaded at runtime or not. default value is false:  a4e3a83

Next i'll add a possibility to install extensions from a mounted directory.

buehner: Maybe you also like to have a look at 4261b1b which introduces the possibility to add custom jars/libs for now.

jodygarnett:
This repository is a place to collaborate, it will be easier once we have a release :)

buehner:
@jodygarnett so you think we should merge this PR to have a better starting point for pull requests?

buehner:
In 82e350ddaa21fdb07276115f01c5ab28e881f73b i enabled the usage of a custom WAR file build from local machine. This is mainly the approach provided by https://github.com/geoserver/docker/pull/2

jodygarnett:
Can we do a 2.19-RC for this? You can release a docker image as early access or something just go get the party started.

buehner:
Of course i can care about this @jodygarnett 
How do you want this to be done from a technical perspective? Just an image "somewhere" on docker hub? Or integrated in some sort of build process?

jodygarnett:
I am just learning about docker:

- There is an osgeo docker repo for test stuff https://repo.osgeo.org/#browse/browse:geoserver-docker
- And I understand we can setup an official geoserver thing on docker hub also?

Please coordinate on the email list!

buehner:
@jodygarnett @randomorder @lpasquali

I also published a version of this approach to docker hub:

`docker run -d --rm -p 8888:8080 --name geoserver terrestris/geoserver:2.19-RC_GSIP-192`

I'll try to find some time until the end of next week to continue the comparison on https://github.com/geoserver/docker/wiki
Next step would be the comparison of the resulting images.

jacobwod:
Great work @buehner! I think that the possibility to set pass `STABLE_EXTENSIONS` to `docker run` is particularly useful in production setups. 🚀 

One thing I can't figure out, however, is this: how can I point Docker to the GS data dir which is located _outside_ the container? I'm aware of `docker build --build-arg GS_DATA_PATH={RELATIVE_PATH_TO_YOUR_GS_DATA}`, but my understanding is that _this affects the container at build time and "bakes in" the data dir_, which is not what I'm looking for. My data dir is basically a directory `/mnt/geoserver_data_dir`, and I'd like to do something like:
```
docker run -d --rm -p 8888:8080 -e STABLE_EXTENSIONS='wps,csw' -e DATA_DIR='/mnt/geoserver_data_dir' --name geoserver terrestris/geoserver 
```

One super-important reason for this approach is the **ability to share data dir between GS instances running in Docker**. Any ideas on how this can be achieved?

jacobwod:
As a reply to my last comment, **binding existing data directory into the container can be easily done**. Here's how I've setup the `terrestris/geoserver:2.19.0` tag in our environment:

```
docker run -it -p 80:8080 -e STABLE_EXTENSIONS='oracle,css' -v /mnt/gisdata/geoserver_data_dir:/opt/geos
ver_data --name geoserver terrestris/geoserver:2.19.0
```

I have mainly one question remaining: **can we find a way to customise Tomcat's `server.xml` within the container?** One use case would be to customise the access log format, which we currently do in `server.xml`:
```
<Valve  className="org.apache.catalina.valves.AccessLogValve" 
  directory="/mnt/gisdata/geoserver_data_dir/logs"
  prefix="access_node" suffix=".log"
  pattern="%{X-Forwarded-For}i %h %l %{X-Control-Header}i %t &quot;%r&quot; %s %b &quot;%{Referer}i&quot; &quot;%{User-Agent}i&quot;"
  renameOnRotate="true" />
```

Dockerfile
```diff
@@ -1,83 +1,87 @@
...
-    DOWNLOAD_EXTENSIONS=false \
-    STABLE_EXTENSIONS='' \
-    DEBIAN_FRONTEND=noninteractive \
-    ADDITIONAL_LIBS_DIR=/opt/additional_libs/ \
-    GEOSERVER_DATA_DIR=/opt/geoserver_data/ \
-    GEOWEBCACHE_CACHE_DIR=/opt/geowebcache_data/
+# Environment variables
+ENV CATALINA_HOME=/opt/apache-tomcat-${TOMCAT_VERSION}
+ENV GEOSERVER_VERSION=$GS_VERSION
+ENV GEOSERVER_DATA_DIR=#######/opt/geoserver_data/
```
buehner, line 19: Why `#######`? ;-)

buehner:
It's working fine now for me. LGTM

weskamm:
@buehner Can we remove the "WIP" from the title please?
@jodygarnett Can this be merged than?

jodygarnett:
Yes you can remove the WIP, and you can merge when you are ready (this is your project to collaborate on).

When you are ready (and a build image is available) you can make a PR to https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/installation/index.rst

aaime:
Given the recent changes, how does it compare with https://github.com/geoserver/docker/pull/2 ?
Have Alessandro/Luca addressed most of the feedback they received?

randomorder:
Hi @aaime
I had a very quick look at this and overall look ok to me. If you don't mind I'would take a closer look next week before merging this PR and closing #2

weskamm:
Still any objections?

randomorder:
nope, fine by me thanks

buehner:
So this is ready for merge now?

PR #2, opened 2021-01-15
WIP: GeoSolutions proposed docker implementation

lpasquali:
This implementation proposed is still to be perfected but it satisfies these initial requests [here](https://github.com/geoserver/geoserver/wiki/GSIP-192):

It supports automated builds from the docker hub if cofigured as per README.md as well custom builds with the provided `custom_build.sh` script


buehner:
Hi, thank you for proposing this alternative approach. I really like the docker multi-stage approach and also the possibility to build with custom/local WAR files, which is a nice feature that is still missing in https://github.com/geoserver/docker/pull/1

I was playing around with your proposal and as you say: This really needs some refinement/fixing. It took me quite a while to get this running in a very truncated version as i got several problems during that process.

As a quickstart i tried

`docker build -t geoserver:test --build-arg GEOSERVER_WEBAPP_SRC="http://sourceforge.net/projects/geoserver/files/GeoServer/2.18.2/geoserver-2.18.2-war.zip" .` 

which is not working. I think the problem is that sourceforge is redirecting such requests and the docker `ADD` is not able to handle this. But in my eyes an "official" geoserver docker solution should also support the "official" geoserver WAR resource.

One more thought on the `ADD` approach: I see the advantage of being flexible regarding the source of the war file (use either local archive or http URL), but docker itself does not recommend the usage of `ADD` due to it's special nature that may lead to confusion and other problems: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
So i think we should try to find a smarter solution on a script basis that equally satisfies this requirement.

Next, i played around with the custom build script approach. I finally got it running in a truncated version. Some lines made problems and i had to adjust or comment them out. I'll add comments on these lines after this post. 

Nonetheless we should definitely provide the possibility to build with custom WARs. So my suggestion for now is that i will pick up the aspect that is still missing in https://github.com/geoserver/docker/pull/1  ("burn a custom WAR file to the image") and try to find a smart solution for that in my approach. (Of course this could also happen with other aspects that may be missing in my approach).

In general, I have the impression that we have two interest groups with regard to the Docker image: One group is primarily interested in the possibilities of the Dockerfile as they want to make use of it in future (without the need for an official pre-built docker image). The other group focuses more on the resulting Docker image, but doesn't want to build it themselves if it is provided by a hub. They just want to pull an existing docker image and configure it for their needs (this is probably the main focus of my approach).

It would be great if we could combine both approaches, so that everybody is happy in the end.

custom_build.sh
```diff
@@ -0,0 +1,191 @@
...
+readonly MARLIN_VERSION=0.9.2
+readonly ARTIFACT_DIRECTORY=./resources
+readonly GEOSERVER_ARTIFACT_DIRECTORY=${ARTIFACT_DIRECTORY}/geoserver/
+readonly DATADIR_ARTIFACT_DIRECTORY=${ARTIFACT_DIRECTORY}/geoserver-datadir/
+readonly PLUGIN_ARTIFACT_DIRECTORY=${ARTIFACT_DIRECTORY}/geoserver-plugins
+readonly FONTS_ARTIFACT_DIRECTORY=${ARTIFACT_DIRECTORY}/fonts/
+readonly MARLIN_ARTIFACT_DIRECTORY=${ARTIFACT_DIRECTORY}/marlin/
+
+function help(){
+	if [ "$#" -ne 6 ] ; then
```
buehner, line 22: This should be `"$#" -ne 5` i think...

Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+        rm ./*zip; \
+    fi \
+    && [ -d "./geoserver" ] || (mkdir -p ./geoserver && unzip ./geoserver.war -d ./geoserver && rm ./geoserver.war)
+
+RUN apt-get update; apt-get upgrade --yes; apt-get install wget --yes
+RUN wget https://downloads.sourceforge.net/project/libjpeg-turbo/1.5.3/libjpeg-turbo-official_1.5.3_amd64.deb && dpkg -i ./libjpeg*.deb && apt-get -f install
+
+WORKDIR /output/plugins
+ARG PLUG_IN_URLS=""
+ADD .placeholder ${PLUG_IN_URLS} /output/plugins/
```
buehner, line 29: This was failing for me and i had to comment it out

Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+# zip files require explicit extracion
+RUN \
+    if [ "${GEOSERVER_WEBAPP_SRC##*.}" = "zip" ]; then \
+        unzip "./*zip"; \
+        rm ./*zip; \
+    fi \
+    && [ -d "./geoserver" ] || (mkdir -p ./geoserver && unzip ./geoserver.war -d ./geoserver && rm ./geoserver.war)
+
+RUN apt-get update; apt-get upgrade --yes; apt-get install wget --yes
+RUN wget https://downloads.sourceforge.net/project/libjpeg-turbo/1.5.3/libjpeg-turbo-official_1.5.3_amd64.deb && dpkg -i ./libjpeg*.deb && apt-get -f install
```
buehner, line 25: Would you recommend to always build in/use the libjpeg-turbo extension and lib?

custom_build.sh
```diff
@@ -0,0 +1,191 @@
...
+
+function download_from_url_to_a_filepath {
+	URL=${1}
+	FILE_PATH=${2}
+	FILE_DOWNLOADED=$(basename "${FILE_PATH}" )
+	if [ -f "${FILE_PATH}" ]; then
+		rm -f "${FILE_PATH}"
+	fi
+	if [ ! -f "${FILE_PATH}" ]; then
+		curl -L "${URL}" --output "${FILE_PATH}"
```
buehner, line 47: This curl download was not working in my case (for plugin downloads). But i could verify that the generated download URLs for the plugins were correct and working in the browser. So i think this could also be related to the way sourceforge is redirecting here...

custom_build.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function download_fonts()  {
+    if [ ! -e "${FONTS_ARTIFACT_DIRECTORY}" ]; then
+        mkdir -p "${FONTS_ARTIFACT_DIRECTORY}"
+    fi
+    download_from_url_to_a_filepath "${EXTRA_FONTS_URL}" "${FONTS_ARTIFACT_DIRECTORY}/fonts.tar.gz"
+}
+
+function download_marlin()  {
```
buehner, line 100: As Marlin renderer is built-in in Java 11 and this function is never called, it could be removed.

custom_build.sh
```diff
@@ -0,0 +1,191 @@
...
+		 .
+}
+
+function main {
+    help ${ALL_PARAMETERS}
+    download_geoserver "${GEOSERVER_VERSION}"
+    clean_up_directory ${PLUGIN_ARTIFACT_DIRECTORY}
+    download_plugin ext monitor
+    download_plugin ext control-flow
+    download_plugin community sec-oauth2-geonode
```
buehner, line 180: i had to comment out all plugin downloads to get this running on my machine

custom_build.sh
```diff
@@ -0,0 +1,191 @@
...
+        if [[ "${PULL_ENABLED}" == "pull" ]]; then
+                DOCKER_BUILD_COMMAND="docker build --pull"
+        else
+                DOCKER_BUILD_COMMAND="docker build"
+        fi;
+	${DOCKER_BUILD_COMMAND} --no-cache \
+		--build-arg GEOSERVER_WEBAPP_SRC=${GEOSERVER_ARTIFACT_DIRECTORY}/geoserver.war \
+    --build-arg PLUG_IN_URLS=$PLUGIN_ARTIFACT_DIRECTORY \
+    --build-arg GEOSERVER_DATA_DIR_SRC=${GEOSERVER_DATA_DIR_DIRECTORY} \
+		-t geosolutionsit/geoserver:"${TAG}-${GEOSERVER_VERSION}" \
```
buehner, line 154: ;-)

buehner:
Update: As stated above, i had problems using a sourceforge URL as source for the war file. 

I now found out that https://downloads.sourceforge.net/project/geoserver/GeoServer/2.18.2/geoserver-2.18.2-war.zip seems to work (instead of http://sourceforge.net/projects/geoserver/files/GeoServer/2.18.2/geoserver-2.18.2-war.zip)

README.md
```diff
@@ -1,2 +1,85 @@
...
+
+## How to build the Dockerfile with no helper scrips
+
+If you want to build the image by yourself just run `docker build` from the root of the repository
+
+` docker build -t geoserver:test . --build-args GEOSERVER_WEBAPP_SRC="./geoserver.war"`
+
+There are [**build arguments**](https://docs.docker.com/engine/reference/commandline/build/) to customize the image:
+- `GEOSERVER_DATA_DIR_SRC` to add your own custom datadir to the final image. This can be a local zip or directory or remote URL (see [ADD](https://docs.docker.com/engine/reference/builder/#add) instruction Doc)
+- `GEOSERVER_WEBAPP_SRC` to add your own custom web app to the final image. This can be a local zip or directory or remote URL (see [ADD](https://docs.docker.com/engine/reference/builder/#add) instruction Doc)
```
buehner, line 35: seems that zip is not extracted automatically by ADD (but tar)

buehner:
In https://github.com/geoserver/docker/commit/82e350ddaa21fdb07276115f01c5ab28e881f73b i enabled the usage of a custom WAR from local machine by using the ADD magic ;-)

custom_build.sh
randomorder: I agree

README.md
randomorder: This is handled by https://github.com/geoserver/docker/pull/2/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R17

custom_build.sh
randomorder: Tag should be updated here

randomorder:
> Hi, thank you for proposing this alternative approach. I really like the docker multi-stage approach and also the possibility to build with custom/local WAR files, which is a nice feature that is still missing in #1
> 
> I was playing around with your proposal and as you say: This really needs some refinement/fixing. It took me quite a while to get this running in a very truncated version as i got several problems during that process.
> 
> As a quickstart i tried
> 
> `docker build -t geoserver:test --build-arg GEOSERVER_WEBAPP_SRC="http://sourceforge.net/projects/geoserver/files/GeoServer/2.18.2/geoserver-2.18.2-war.zip" .`
> 
> which is not working. I think the problem is that sourceforge is redirecting such requests and the docker `ADD` is not able to handle this. But in my eyes an "official" geoserver docker solution should also support the "official" geoserver WAR resource.
> 
> One more thought on the `ADD` approach: I see the advantage of being flexible regarding the source of the war file (use either local archive or http URL), but docker itself does not recommend the usage of `ADD` due to it's special nature that may lead to confusion and other problems: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
> So i think we should try to find a smarter solution on a script basis that equally satisfies this requirement.
> 
> Next, i played around with the custom build script approach. I finally got it running in a truncated version. Some lines made problems and i had to adjust or comment them out. I'll add comments on these lines after this post.
> 
> Nonetheless we should definitely provide the possibility to build with custom WARs. So my suggestion for now is that i will pick up the aspect that is still missing in #1 ("burn a custom WAR file to the image") and try to find a smart solution for that in my approach. (Of course this could also happen with other aspects that may be missing in my approach).
> 
> In general, I have the impression that we have two interest groups with regard to the Docker image: One group is primarily interested in the possibilities of the Dockerfile as they want to make use of it in future (without the need for an official pre-built docker image). The other group focuses more on the resulting Docker image, but doesn't want to build it themselves if it is provided by a hub. They just want to pull an existing docker image and configure it for their needs (this is probably the main focus of my approach).
> 
> It would be great if we could combine both approaches, so that everybody is happy in the end.

Thanks a lot for your review @buehner

> It would be great if we could combine both approaches, so that everybody is happy in the end.

I agree. It should be relatively easy to combine both approached and get the best out of the two

custom_build.sh
randomorder: I confirm I have similar behaviour where the final war.zip for geoserver is instead an HTMLpage @lpasquali
```
/output/webapp# head geoserver-2.18.2-war.zip
<!doctype html>
<!-- Server: sfs-consume-6 -->
<html class="no-js" lang="en">
    <head>
        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />


        <script>
    var __gdpr = true;
...
```

lpasquali:
> `docker build -t geoserver:test --build-arg GEOSERVER_WEBAPP_SRC="http://sourceforge.net/projects/geoserver/files/GeoServer/2.18.2/geoserver-2.18.2-war.zip" .`

Hello @buehner many thanks for your comments! 

I will start from the first one: unfortunately that url not working for you it is a limitation of sourceforge, taking from the custom_build.sh script, you could use a direct mirror link linke this one "https://netcologne.dl.sourceforge.net/project/geoserver/GeoServer/2.18.1/geoserver-2.18.1-war.zip"

the Dockerfile build would had failed anyway because it missed a file in the context "run_tests.sh" which I just added.
this command should do then the trick:

```
docker build -t geoserver:test --build-arg GEOSERVER_WEBAPP_SRC="https://netcologne.dl.sourceforge.net/project/geoserver/GeoServer/2.18.1/geoserver-2.18.1-war.zip" .
```

custom_build.sh
lpasquali: yes it is correct

Dockerfile
lpasquali: was it failing with custom_build.sh?

custom_build.sh
lpasquali: @buehner how was the url this time?

custom_build.sh
lpasquali: removed

custom_build.sh
lpasquali: they were there just as examples but they should work, I'd need output of:
```
bash -x ./custom_build.sh <options>
```
to understand well

custom_build.sh
lpasquali: done

Dockerfile
randomorder: I'd remove all extensions from the official images by default

jodygarnett:
Can we do a 2.19-RC for this? You can release a docker image as early access or something just go get the party started.

randomorder:
Yes, no problem @jodygarnett 
We'll make the image available before the end of the week

randomorder:
Hi @jodygarnett @buehner 

We pushed the image on GeoSolutions org repo on the docker hub:
`docker run -d --rm -p 8888:8080 --name geoserver geosolutionsit/geoserver:2.19-RC_GSIP-192`

The image was built according to the instr. in the readme. We used the default datadir that comes with the 
`docker build -t geosolutionsit/geoserver:2.19-RC_GSIP-192 --build-arg GEOSERVER_WEBAPP_SRC=./resources/geoserver/geoserver.war .`

In the process I noticed in the README there are no instr. on how to externalize the datadir. @lpasquali is amending it right now.

PR #5, opened 2022-08-12
Don't try to remove non-existent directories

mprins:


buehner:
Thank you. I will merge

PR #9, opened 2022-09-05
Skip demo data on startup

JakobMiksch:
- This PR makes it possible to create an empty GeoServer without demo data, it can be controlled via `SKIP_DEMO_DATA=true`
- The motivation is have a clean GeoServer in production scenarios
- the current behavior of the image stays the same i.e. a new created GeoServer will be populated with demo data
- The change of the PR tries to build around the current usage of the environment variable [`GEOSERVER_REQUIRE_FILE`](https://docs.geoserver.org/stable/en/user/datadirectory/setting.html#require-files-to-exist). This variable is also checked by GeoServer itself and is used to check if demo data should be created. 

Test setup: 
```shell
# build image
docker build -t gs-dev . 

# creates a "clean" GeoServer without demo data
docker run \
  -p 8080:8080 \
  -e SKIP_DEMO_DATA=true \
  gs-dev
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ To pull an official image use ``docker.osgeo.org/geoserver:{{VERSION}}``, e.g.:
 docker pull docker.osgeo.org/geoserver:2.21.1
 ```
 
-Afterwards you can run the pulled image locally with: 
+Afterwards you can run the pulled image locally with:
 
 ```
 docker run -it -p 80:8080 docker.osgeo.org/geoserver:2.21.1
@@ -48,11 +48,16 @@ To use an external folder as your geoserver data directory.
 ```
 docker run -it -p 80:8080 \
   --mount src="/absolute/path/on/host",target=/opt/geoserver_data/,type=bind \
-  docker.osgeo.org/geoserver:2.21.1 
+  docker.osgeo.org/geoserver:2.21.1
 ```
 
 An empty data directory will be populated on first use. You can easily update GeoServer while
 using the same data directory.
+
+### How to start a GeoServer without sample data?
+
+This image populates GeoServer with demo data by default. For production scenarios this is typically not desired.
+Setting the environment varibale `SKIP_DEMO_DATA` can be set to `true` to create an empty GeoServer.
 
 ### How to download and install additional extensions on startup?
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -48,26 +48,31 @@ To use an external folder as your geoserver data directory.
...
+  docker.osgeo.org/geoserver:2.21.1
 ```
 
 An empty data directory will be populated on first use. You can easily update GeoServer while
 using the same data directory.
 
+### How to start a GeoServer without sample data?
+
+This image populates GeoServer with demo data by default. For production scenarios this is typically not desired.
+Setting the environment varibale `SKIP_DEMO_DATA` can be set to `true` to create an empty GeoServer.
```
marcjansen, line 60:
```suggestion
Setting the environment variable `SKIP_DEMO_DATA` can be set to `true` to create an empty GeoServer.
```

marcjansen (commented):
One minor comment and a question. Wouldn't it be enough i one mounted a specific geoserver data directory? Empty or not? I think for most production settings this would be a sane assumption.

buehner:
I have the same question: Wouldn't you just mount your data directory in a custom data directory and then the demo data are not relevant anymore?

chrismayer:
@buehner @marcjansen Totally right, mounting a custom (empty) data directory would achieve the same. The idea is more to not copying around unnecessary files, which are unused. Since skipping the demo data is optional, so the current behavior is not changed (you get the demo data if doing anything) I assume this does not hurt and maybe makes things easier in some cases (at least to our experience). In case you do not agree, that's totally fine. Then maybe a short hint on the README or somewhere else would be helpful howto start if you need a clean setup.

buehner:
As you say, merging this would not change anything by default, so i don't have a problem to merge it, if this may help in certain scenarios. Just wanted to understand the intention ;-) 

I would suggest to wait a bit (maybe someone else still likes to comment on this) and if there are no concerns, i can merge it within the next days.

README.md
```diff
@@ -48,26 +48,31 @@ To use an external folder as your geoserver data directory.
...
+  docker.osgeo.org/geoserver:2.21.1
 ```
 
 An empty data directory will be populated on first use. You can easily update GeoServer while
 using the same data directory.
 
+### How to start a GeoServer without sample data?
+
+This image populates GeoServer with demo data by default. For production scenarios this is typically not desired.
+Setting the environment variable `SKIP_DEMO_DATA` can be set to `true` to create an empty GeoServer.
```
marcjansen, line 60:
```suggestion
The environment variable `SKIP_DEMO_DATA` can be set to `true` to create an empty GeoServer.
```

marcjansen:
One more wording suggestion otherwise I'll leave the rest to the experts.

weskamm (approved):

PR #11, opened 2022-09-29
Add support to change Tomcat default HTTP port

williamccondori:


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:22.04
 
 # The GS_VERSION argument could be used like this to overwrite the default:
 # docker build --build-arg GS_VERSION=2.21.2 -t geoserver:2.21.2 .
-ARG TOMCAT_VERSION=9.0.65
+ARG TOMCAT_VERSION=9.0.67
```
mprins, line 5: See also #12 which updates to the current release

mprins:
Not sure why anyone would want this given that geoserver provides proxy url and docker can map to any port.
Also with port 80 you would need to run tomcat as root which is bad practice

buehner:
As @mprins said, docker can map to any port, so i don't think we need that. Maybe you can explain your scenario, if you think that this is still needed?

PR #14, opened 2022-11-07
Notes how to include a custom extension

JakobMiksch:
- Notes how to include a custom extension
- Emphasize that it is necessary to mount jar files and no zip packages


buehner:
Thx. I will merge

PR #17, opened 2023-01-25
Version 2.22.0 in docker compose and readme

herrapa:
Keep geoserver version in Docker compose demo file and readme in sync with Dockerfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,19 +18,19 @@ This Dockerfile can be used to create images for all geoserver versions since 2.
 To pull an official image use ``docker.osgeo.org/geoserver:{{VERSION}}``, e.g.:
 
 ```shell
-docker pull docker.osgeo.org/geoserver:2.21.1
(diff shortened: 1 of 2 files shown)


JakobMiksch:
Good catch. The version `2.22.0` is not published on the registry yet: https://docker.osgeo.org/#browse/browse:geoserver-docker:v2%2Fgeoserver%2Ftags
Ideally the PR will be merged afterwards

buehner:
Yes. I can wait with merging until 2.22.0 is published

weskamm:
I think we should change the docs to use `latest` tag instead of fixing version numbers after every release.
But before that, we need the latest tag, as it currently does not exist yet

buehner:
I will merge now as 2.22.0 is available: https://docker.osgeo.org/#browse/browse:geoserver-docker:v2%2Fgeoserver%2Ftags%2F2.22.0

PR #18, opened 2023-02-01
Adding build and release script

weskamm:
This adds an release script to build and publish the docker image for a given version.

The variables `DOCKERUSER` and `DOCKERPASSWORD` have to be set with valid credentials before this script can push the image to the osgeo repo.

You need to pass the version as first and the type as second argument, where type has to be one of `build`, `publish` or `buildandpublish`.

Examples:

`./release.sh 2.22.1 build`

`./release.sh 2.22.0 publish`

`./release.sh 2.22.1 buildandpublish`


buehner:
Nice! Thx. Maybe it makes sense to add the usage instructions also to the README?

weskamm:
Well as this is for CI purposes only and it has a description in this PR and the script itself i think its not needed

dnlkoch:
We should also consider to add the file to the `.dockerignore`.

buehner:
Thx a lot. I will merge now as this will only make the script available, but not change anything on Docker side. Anything else can be solved in follow up PRs.

jodygarnett:
I started looking at the Jenkins configuration and quickly decided I did not know how to responsibly manage the credentials.

I was hoping to see one of the CITE examples to learn from but could not find one that deploys to osgeo repo.

tbarsballe:
Most Jenkins credentials are managed through the Credentials section (Dashboard > Manage Jenkins > Credentials) and added to jobs in the Build Environment section.

It looks like this is not the case for the CITE examples though - those appear to go through a managed Maven settings.xml that contains credentials. This is set up through the Managed Files instead (Dashboard > Manage Jenkins > Managed files), though like credentials, is added to the job in the Build Environment section.

Looking at how this deployment implementation is set up, it seems like adding a new Credential for docker deploy is most appropriate - specifically a "Username with Password" credential. These can then be mapped to environment variables in the job configuration on Jenkins.

I can help set this up if that'd be easiest. @jodygarnett - should we just be using the same credentials as the CITE deploy uses? I can pull those from the settings.xml

jodygarnett:
Yes that would be great, thank you.

tbarsballe:
Alright. I've added the OSgeo Docker repo as a "Username with Password" credential.
I'm not seeing a build for geoserver/docker yet to attach this to - has one been created yet?

jodygarnett:
I created geoserver-release-docker job but was not able tog eat it to check out geoserver/docker repo.  Think the git user may need permissions.

tbarsballe:
Alright, I can take a look at that job and set up credentials accordingly. Jenkins seems to be down at the moment (it was up 30 mins ago, not sure whats going on), but I'll let you know when I've had a change to update the job.

weskamm:
Great news, so it does now build automatically on new release?

jodygarnett:
I can go and add that something to do after geoserver-release-publish ...

buehner:
Great. Thx a lot. Would it be possible to also tag the latest succesfull build with the `latest` tag as suggested here: https://github.com/geoserver/docker/pull/17#issuecomment-1410000929 ?

weskamm:
Seems like it currently does not build automatically on release, as the docker image for 2.23.0 is still missing.
Maybe you could have another look @jodygarnett ?

PR #20, opened 2023-03-25
Making sure additional fonts are correctly counted

dnavre:
Currently the additional fonts installation fails as they are wrongly counted in the startup.sh script.


buehner:
Thx for this. I will merge

PR #23, opened 2023-06-07
[GEOS-11016] Use multi-stage build to avoid caching snapshot download

jodygarnett:
Docker is caching the download step and is not noticing when new war is available for download. This leads to the **geoserver.war** being out of sync with the extensions and community modules installed.

The Dockerfile has three stages:
- tomcat
- download
- install

For a nightly build run with ``--no-cache-filter '!tomcat'`` to avoid caching download and install stages.


jodygarnett:
Local testing of ``build/release.sh`` is good.

As this changes needs to be tested at the same time as build server job [geoserver-release-docker](https://build.geoserver.org/view/release/job/geoserver-release-docker/) I am going to merge and test.

PR #24, opened 2023-07-07
org.marlin.pisces to sun.java2d.marlin 

zionio:
Hi, 
 as of https://github.com/bourgesl/marlin-renderer/releases isn't it better to use 
`-Dsun.java2d.renderer=sun.java2d.marlin.DMarlinRenderingEngine` with openjdk11 ?

 ref: https://www.geocat.net/docs/geoserver-enterprise/2020.5/install/production/marlin.html#check-rendering-engine

Java | Rendering Engine | Recommendation
-- | -- | --
Java 8 | sun.dc.DuctusRenderingEngine | This is the Java 8 default rendering engine, recommend installing Marlin.
Java 8 | org.marlin.pisces.MarlinRenderingEngine | Correctly configured with Marlin Rendering Engine
Java 11 | sun.java2d.marlin.DMarlinRenderingEngine | Java 11 defaults to MarlinRenderingEngine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ENV CATALINA_OPTS="\$EXTRA_JAVA_OPTS \
     -Djavax.servlet.response.encoding=UTF-8 \
     -D-XX:SoftRefLRUPolicyMSPerMB=36000 \
     -Xbootclasspath/a:$CATALINA_HOME/lib/marlin.jar \
-    -Dsun.java2d.renderer=org.marlin.pisces.PiscesRenderingEngine \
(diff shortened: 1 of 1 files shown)


buehner:
You are right. Good catch! Don't know why it was still configured to Pisces. I will merge. Thank you

aaime:
If you are using the Java built-in renderer, you can simplify and just remove the setting then. The GeoServer stand-alone Java binary uses an explicit setting to force usage of a more recent version of Marlin, than the one shipped with JDK 11.

PR #26, opened 2023-07-11
Allow setting proxy base URL in environment

sgrimm:
When bringing GeoServer up behind a load balancer or proxy that does HTTPS termination, e.g., using Kubernetes or Amazon ECS, there's a bootstrapping problem: you can't log in as the admin to set the proxy URL until the proxy URL is already set because GeoServer thinks its URL is HTTP, not HTTPS.

Add support for configuring the proxy URL in `web.xml` using an environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,13 @@ The environment variable `ROOT_WEBAPP_REDIRECT` can be set to `true` to issue a 
 
 The ``startup.sh`` script allows some customization on startup:
 
-* ``INSTALL_EXTENSIONS`` to ``true`` to download and install extensions
-* ``STABLE_EXTENSIONS`` list of extensions to download and install
-* ``CORS_ENABLED``
+* ``INSTALL_EXTENSIONS`` to ``true`` to download and install extensions.
+* ``STABLE_EXTENSIONS`` list of extensions to download and install.
+* ``CORS_ENABLED`` to ``true`` to enable CORS support. The following environment variables can be used to customize the CORS configuration.
+  * ``CORS_ALLOWED_ORIGINS`` (default ``*``)
+  * ``CORS_ALLOWED_METHODS`` (default ``GET,POST,PUT,DELETE,HEAD,OPTIONS``)
+  * ``CORS_ALLOWED_HEADERS`` (default ``*``)
+* ``PROXY_BASE_URL`` to the base URL of the GeoServer web app if GeoServer is behind a proxy. Example: ``https://example.com/geoserver``.
 
 Example installing wps and ysld extensions:
 
--- a/startup.sh
+++ b/startup.sh
@@ -1,6 +1,8 @@
 #!/bin/sh
 echo "Welcome to GeoServer $GEOSERVER_VERSION"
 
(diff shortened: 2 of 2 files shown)


jodygarnett:
Yes, most of the things like PROXY_BASE_URL need to be set from a docker image are available as environmental variables.

reinout:
I've tried it with setting the `PROXY_BASE_URL` variable: yes, works fine.

There's one difference to my `X-Forwarded-For/X-Forwarded-Proto` solution, though. `X-Forwarded-For` picks up the hostname+protocol and also automatically works fine with adding `/geoserver` to the urls.
With `PROXY_BASE_URL`, you need to explicitly add `/geoserver`. I originally set just `https://my.server.net` as base url, but that gave me 404's. Setting it to `https://my.server.net/geoserver` worked. 

For this Docker it will be okay as there will only be one geoserver inside the tomcat instance. But in general, if there are more webapps installed inside tomcat, that one proxy base url won't be right for all of them. That's why I didn't expect the /geoserver path part.

Anyway... the env var works. Perhaps we need to document it somewhere?

reinout:
Correction, I *thought* that the env var worked. I've since gone back to my original solution.

The problem was that pages and map layers would load fine, but some UI elements didn't work, like the pagination on the map preview screen. Clicking just didn't work and gave no reaction, I couldn't see anything in my web dev network tab.

I've done another test this afternoon. I suspected I might have to add a trailing slash to `PROXY_BASE_URL`. That helped to get the UI to at least result in actual requests. They would return a `400 bad request` with the message `Origin does not correspond to request`. So: csrf problems. I saw some hints to disable csrf (=bad) and hints to add the domain to a whiltelist. And I saw hints that suggested http/https wasn't getting properly detected.

So.... back to the solution I got from https://github.com/docker-library/tomcat/issues/107#issuecomment-639034758 .

nickgraz:
I second that this works and is needed. Without this, setting `PROXY_BASE_URL` with env var _seems_ to work, but some of the forms that you access after logging in don't work. For example adding a layer from **Data -> Layers -> Add a new layer**, when selecting something from the dropdown I get a 403. After implementing this pull request, it works as expected.

nickgraz:
Ignore my previous comment, the issue I was facing was actually due to an incorrect value for `CORS_ALLOWED_ORIGINS`. Sorry about that! It does work as expected with the OOB `PROXY_BASE_URL` env var in my case.

buehner:
@reinout @nickgraz @sgrimm 

What about this? I am a bit confused and not sure whether this is needed or not? In case we need it, an update would be great as we have merge conflicts now. Thx

nickgraz:
I don't think that this one is needed anymore. Setting `PROXY_BASE_URL` as an environment variable works as expected out of the box, without this pull request.

sgrimm:
Agreed, this isn't needed any more. When I made this PR, I didn't know about `PROXY_BASE_URL`, so if there's an issue here, it's maybe a documentation problem rather than a code problem.

buehner:
Thank you, i will close the PR.

PR #27, opened 2023-07-18
How about change base image to Alpine?

alapierre:
image based on `ubuntu 22.0` has significant number of security problems:

```shell
trivy image geoserver-docker.osgeo.org/geoserver:2.23.1
```

```
geoserver-docker.osgeo.org/geoserver:2.23.1 (ubuntu 22.04)

Total: 104 (UNKNOWN: 0, LOW: 41, MEDIUM: 63, HIGH: 0, CRITICAL: 0)
```

When you change base image to security patched Alpine it will looks like that:

```shell
trivy image lapierre/geoserver:2.22.4
```

```
lapierre/geoserver:2.22.4 (alpine 3.18.2)

Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
```

Of course, there will still be vulnerabilities in Java libraries that cannot be eliminated by simply changing the base image.

```
Java (jar)

Total: 16 (UNKNOWN: 2, LOW: 0, MEDIUM: 3, HIGH: 6, CRITICAL: 5)
```

Changes in this PR:

- change image to security patched Java Alpine;
- add Makefile and .env for easier build and release


buehner:
Thank you. I generally like the idea of small footprints, but I know that there were problems with alpine in the past when trying to use extended features. In our company based geoserver image for example, we started years ago with alpine and then had to switch back to an ubuntu base image as there were problems with including the `gdal` or `proj` libraries as you can see [here](https://github.com/terrestris/docker-geoserver/pull/21). I don't know whether these problems still exist, but this could be tested, i guess.

Another "contra" would be: Changing the base image like this could cause problems for people that are already using the published ubuntu based-image and extend it for their specific reasons by using the FROM clause. I just don't know how many people would be affected by such a fundamental change.

I'd also like to know the opinion of some core developers. Maybe there are other pro/con arguments regarding the base image? @aaime @jodygarnett

alapierre:
many publicly available Docker images exist in more than one variant - eg. PostgreSQL official image is based on Alpine or Ubuntu, redis, etc. Than users can choose what is more important to them - security and slow footprint or more  flexibility wit bigger and less secure base image.  

We are testing Alpine based Geoserver in production in this moment and I can maintain Alpine fork.

jodygarnett:
I would like to see how you do that? Can we setup multiple supported things here?
(I am asking as I would like to have an M1 image to work from)

alapierre:
yes, it is no problem to support multi variant image in one git repo

buehner:
I like the idea to support multiple image variants (ubuntu or alpine based). Do you think you could extend this PR to prepare something like this @alapierre ?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04 as tomcat
+FROM lapierre/java-alpine:11 as tomcat
```
mprins, line 1: why not use a Docker vetted "Official image"? instead of pushing your own JDK version (having a full JDK is total overkill)
```suggestion
FROM eclipse-temurin:11-jre-alpine as tomcat
```

buehner:
I agree with @mprins that an approach with official images should be preferred.

@alapierre Do you plan to continue on this or can this be closed?

alapierre:
Hey, Yes I'm Planning but curentyly can't find enough time.

buehner:
I close this as nothing happened for months. We can re-open if this is a topic again

PR #28, opened 2023-08-30
Add support for `access-control-allow-credentials` CORS header when `CORS_ALLOWED_ORIGINS` is not a wildcard

nickgraz:
This is needed to access the internal GeoServer using CORS from ArcGIS since the requests are sent _with credentials_. 

If the `CORS_ALLOWED_ORIGINS` is not supplied (defaults to `*`), or is supplied with `*`, it will set the value to `false`, which is the [default Tomcat behavior](https://github.com/apache/tomcat/blob/66261e42e5f188d42a5e02632d5b9bcacc2a9903/java/org/apache/catalina/filters/CorsFilter.java#L955).


reinout:
When I look at the tomcat link you provide, tomcat sets `DEFAULT_SUPPORTS_CREDENTIALS` to false. Which seems like the proper default for CORS: unless explicitly enabled, cross-site credentials is the number one thing to block.

If the `CORS_ALLOWED_ORIGINS` is not the default `*`, it means someone has been extra careful to restrict CORS access. Your pull requests "rewards" that careful behaviour with also enabling credentials. That doesn't sound right to me.

Wouldn't a separate setting to allow credentials be handier? Independent of the allowed origins?

nickgraz:
Great point. I updated the PR to reflect that request. `CORS_ALLOW_CREDENTIALS` is now and env var. It still only applies if it is true _and_  `CORS_ALLOWED_ORIGINS` is not the default. 

I also updated the readme with that info, and the other CORS env vars (from this [PR](https://github.com/geoserver/docker/pull/26)) and the `PROXY_BASE_URL` (based on [this](https://github.com/geoserver/docker/pull/26#issuecomment-1648275258) comment).

buehner:
Thx for the contribution. Looks good to me, so I will merge.

PR #29, opened 2023-09-30
Handle Release Candidate branches

petersmythe:
As per email: `[Geoserver-users] GeoServer 2.24-RC released. Please help us test it`, Andrea says:

```
If I remember correctly, we have had some issue with the docker release script
not handling RC correctly? Like, it's not doing the right thing when it sees "2.24-RC" in the version
and ended up tagging it as part of 2.24.x instead (which I assume has been overwritten by a proper
nightly build by now).
```
This PR adds support for the `-RC` branches.

The publish mode has not been tested due to lack of geoserver-docker.osgeo.org credentials

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -9,7 +9,7 @@ function usage() {
   echo "$0 <mode> <version> [<build>]"
   echo ""
   echo " mode : The mode. Choose one of 'build', 'publish' or 'buildandpublish'"
-  echo " version : The released version to build an docker image for (eg: 2.23.1, ${MAIN}-SNAPSHOT)"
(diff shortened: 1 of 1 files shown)


petersmythe:
@jodygarnett, could you please review?

jodygarnett:
Thank you very much for this @petersmythe, I guess we can use it to retroactively make a 2.24-RC image.

jodygarnett (approved):

jodygarnett:
It appears to have worked: https://build.geoserver.org/view/release/job/geoserver-release-docker/396/

Anyone want to try?

PR #31, opened 2023-10-26
Added support for a PostgreSQL JNDI resource.

mbosecke:
This pull request adds support for creating a PostgreSQL JNDI resource in Apache Tomcat. 

- Included a blurb on the README.md on how to use.
- Provides a default "context.xml" that a user can easily configure using environment variables.
- Provides a mechanism for the user to override the default "context.xml" with their own version of the file for more advanced customization of the connection pool.

Heavily inspired by the work done at https://github.com/kartoza/docker-geoserver.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ ENV CATALINA_OPTS="\$EXTRA_JAVA_OPTS \
 # init
 RUN apt update \
 && apt -y upgrade \
-&& apt install -y --no-install-recommends openssl unzip gdal-bin wget curl openjdk-11-jdk \
+&& apt install -y --no-install-recommends openssl unzip gdal-bin wget curl openjdk-11-jdk gettext \
 && apt clean \
 && rm -rf /var/cache/apt/* \
 && rm -rf /var/lib/apt/lists/*
@@ -92,6 +92,9 @@ ENV ADDITIONAL_LIBS_DIR=/opt/additional_libs/
 ENV ADDITIONAL_FONTS_DIR=/opt/additional_fonts/
 ENV SKIP_DEMO_DATA=false
 ENV ROOT_WEBAPP_REDIRECT=false
+ENV POSTGRES_JNDI_ENABLED=false
+ENV CONFIG_DIR=/opt/config
+ENV CONFIG_OVERRIDES_DIR=/opt/config_overrides
 
 EXPOSE 8080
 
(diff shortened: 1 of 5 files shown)


buehner:
Thank you for the contribution. I really like the JNDI feature and would like to merge it, but I think we still need some clarification.

I tested your changes/branch in a local setup by using the following content for the `docker-compose-demo.yml`. I think you should rebase against the latest master (to avoid version problems, see https://github.com/geoserver/docker/pull/33 )

```
version: '3'
services:
  geoserver:
    build:
      context: .
      args:
        - GEOSERVER_VERSION=2.24.0
        - CORS_ENABLED=true
        - CORS_ALLOWED_METHODS=GET,POST,PUT,HEAD,OPTIONS
    ports:
      - 80:8080
    environment:
      - INSTALL_EXTENSIONS=true
      - STABLE_EXTENSIONS=wps,csw
      - EXTRA_JAVA_OPTS=-Xms1G -Xmx2G
      - POSTGRES_JNDI_ENABLED=true
      - POSTGRES_HOST=postgis
      - POSTGRES_PORT=5432
      - POSTGRES_DB=geoserver
      - POSTGRES_USERNAME=geoserver
      - POSTGRES_PASSWORD=geoserver
      - POSTGRES_JNDI_RESOURCE_NAME=java:comp/env/jdbc/postgres
    volumes:
      - ./geoserver_data:/opt/geoserver_data/:Z
      - ./additional_libs:/opt/additional_libs:Z # by mounting this we can install libs from host on startup
  postgis:
      image: postgis/postgis:16-3.4-alpine
      ports:
        - 5555:5432
      environment:
        POSTGRES_USER: geoserver
        POSTGRES_PASSWORD: geoserver
      volumes:
        - ./postgis/postgresql_data:/var/lib/postgresql/data:Z

```

Once this is started with `docker-compose -f docker-compose-demo.yml up --build`, I can connect to the database and I can login to GeoServer.

But when I try to add a JNDI datastore in GeoServer, I get this error: `Error creating data store, check the parameters. Error message: Cannot find JNDI data source: java:comp/env/jdbc/postgres`

So I connected into the running geoserver container and `cat /opt/config/context.xml` printed the file in its original version (as provided in this MR). It seems to me that the environment variables have not been substituted. But I can confirm they exist by echoing them (e.g. `echo $POSTGRES_JNDI_RESOURCE_NAME` ).

So I guess that the `envsubst` in the `startup.sh` is not working as expected. Can you confirm this? Or am I missing something?

mbosecke:
Thanks for the review! I also struggled with that docker-compose-demo but I think I have it resolved now and I learnt a few things along the way. 

The root of the issue is that in Apache tomcat we have to name the resource "jdbc/postgres" but in Geoserver we have to refer to the resource as "java:comp/env/jdbc/postgres", so there's a bit of a confusing discrepancy there. 

I've fixed the default value specified in "startup.sh" and I adjusted the README file to mention how to use this properly. 

I also committed your changes to the docker-compose-demo because I figured there's some value in having a full-featured demo but let me know if that's now what you'd prefer and I can revert that back. 

Lastly, I rebased onto master.

Let me know if it's still not working for you.

mbosecke:
In regards to the variable substitution, I think there's just a bit of confusion there. The geoserver container will have up to three versions of the "context.xml" file:

1. `/opt/config/context.xml`: A staging version of the context.xml prior to variables being substituted that is provided by this image.
2. `/opt/config_overrides/context.xml`: An optional staging version of the context.xml that the end-users can provide to override our default, which also does _not_ yet have the variables substituted.
3. `/opt/apache-tomcat-9.0.82/conf/context.xml`: The final version with substituted variables. 

So you should be able to see the final version using:

`cat /opt/apache-tomcat-9.0.82/conf/context.xml `

Hopefully that makes sense.

buehner:
Thx a lot for the explanations. I could successfully start and verify this and I also like the docker-compose-demo being extended.

I will merge now.

petersmythe:
This PR works nicely for me, using the customized "context.xml" with multiple JNDI connections.  Thank you @mbosecke for the contribution!

PR #32, opened 2023-11-08
Implement EXPOSE port

petersmythe:
See https://docs.docker.com/engine/reference/builder/#expose

Otherwise Docker Desktop (Windows) does not know that any ports are exposed:

image


buehner:
Thank you, I will merge

PR #33, opened 2023-11-09
fix: Use consistent version number in Dockerfile

buehner:
Due to the use of the multistage build feature, we have multiple declarations of `GS_VERSION` in the Dockerfile. For some reason, this value diverged during the latest releases. This MR just fixes the value, but we should investigate the source of the problem and fix it in a follow up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,7 @@ RUN apt purge -y  \
 
 FROM tomcat as download
 
-ARG GS_VERSION=2.23.1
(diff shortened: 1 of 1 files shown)


buehner:
Maybe this can help to find a better ARG handling in future: https://github.com/moby/moby/issues/37622#issuecomment-412101935

reinout:
Yes, it looks like you now only would need a single `ARG GS_VERSION=2.24.0` at the top of the file, before the first step.

PR #35, opened 2023-12-01
Docs: Add docker ENV vars in README

jashanbhullar:
Adding all the env vars in table format in README. The list was obtained using `docker inspect IMAGE_NAME`. 
Some of the values are duplicated with some examples above. 
Some points that can be further added are:
- Reference to docs for JAVA variables 
- Reference to dosc for CATALINA variables 
- Some variables should not be changed like `GEOSERVER_VERSION` so maybe mark them as `should not change`? (confused on this) 
- Add sample value for empty places

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ docker-compose -f docker-compose-demo.yml up --build
 ```
 ## Environment Variables
 
-Following is the list of the all the environment variables that can be passed down to the geoserver docker image, you can check the defaul values for an image using `docker inspect [IMAGE_NAME]`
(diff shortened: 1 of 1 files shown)


jodygarnett:
For JAVA_OPTS and CATALINA_OPTS the only thing we can do is link to the Java command line options for Java 11 (which is the version of Java we are running).

These two environmental variables are added to the java command line that runs tomcat application server. 
The `-X` options are for the Java Virtiable Machine. The `-D` are java system properties can be for tomcat or geoserver and geotools and ... so I do not know an easy way to provide a guide?

It is hard to find this information online because oracle docs are no longer listed for Java 11.

jashanbhullar:
Thanks @jodygarnett , I have addressed the reconfigurable params issue. This PR is ready to be merged.

PR #36, opened 2023-12-03
Update docker-compose-demo.yml

davidce:
postgis service map and expose the port to 5555


buehner:
I think you are mixing things up here. `POSTGRES_PORT` configures the JNDI connection (see https://github.com/geoserver/docker/pull/31) within the containers/internal docker network (tomcat/geoserver -> database), so port `5432` is correct here to have this working as expected. The database instance itself is already exposed on `5555` in the demo setup.

PR #38, opened 2023-12-11
Introduce health checks for GeoServer (and postgis db in demo)

ahennr:
This PR introduces a configurable docker health check (see also here).
Currently, the logo of the landing page is used for check, however it can be configured using the HEALTHCHECK_URL environment variable.

In the demo docker compose file a corresponding block for the postgis database has been added.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,6 +95,7 @@ ENV ROOT_WEBAPP_REDIRECT=false
 ENV POSTGRES_JNDI_ENABLED=false
 ENV CONFIG_DIR=/opt/config
 ENV CONFIG_OVERRIDES_DIR=/opt/config_overrides
+ENV HEALTHCHECK_URL=http://localhost:8080/geoserver/web/wicket/resource/org.geoserver.web.GeoServerBasePage/img/logo.png
 
 EXPOSE 8080
 
@@ -126,3 +127,6 @@ RUN chmod +x /opt/*.sh
 ENTRYPOINT ["/opt/startup.sh"]
 
 WORKDIR /opt
+
+HEALTHCHECK --interval=1m --timeout=20s --retries=3 \
+  CMD curl --fail $HEALTHCHECK_URL || exit 1
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ Following is the list of the all the environment variables that can be passed do
 | ADDITIONAL_FONTS_DIR | Sets the directory for additional fonts used by GeoServer | `/opt/additional_fonts/` |
 | SKIP_DEMO_DATA | Indicates whether to skip the installation of demo data provided by GeoServer | `false` |
 | ROOT_WEBAPP_REDIRECT | Indicates whether to issue a permanent redirect to the web interface | `false` |
+| HEALTHCHECK_URL | URL to the resource / endpoint used for `docker` health checks | `http://localhost:8080/geoserver/web/wicket/resource/org.geoserver.web.GeoServerBasePage/img/logo.png` |
 
 The following values cannot really be safely changed (as they are used to download extensions and community modules as the docker image first starts up).
 | VAR NAME | DESCRIPTION | SAMPLE VALUE |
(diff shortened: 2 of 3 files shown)


reinout:
A health check is a really good idea. I've also got one in my customisations on top of the official dockerfile:

    HEALTHCHECK --interval=20s --timeout=10s --retries=3 --start-period=150s \
        CMD curl --fail --max-time 3 http://localhost:8080/geoserver/web/ || exit 1

There are a couple of differences with your PR. You make the URL configurable: good idea!

A **second difference** is the default URL. You're checking a small image and I'm checking the main web page. A small, inexpensive URL is generally a better idea. I've chosen the main web page as for me geoserver often takes quite some time to start up, having the health check report its healthiness when it is really ready seemed like a good idea to me. (But this is up for debate, of course).

I'm a bit concerned that the main web page is too heavy. But your image check seems like something that will be handled by some static file hosting part of the code: will that fail to load when there's a serious database connection issue? I haven't investigated.

Does anyone know of a proper `/geoserver/weg/health` url? I couldn't find anything when I looked for it.

A **third difference** is the interval and timeout settings. Depending on the nature of the URL, you want different settings. As I use the main webpage, I really needed `--start-period=150s` to allow geoserver time to startup. Otherwise the container gets killed because of failing checks before it got time to start up. But that's also because I'm checking more often. You check once a minute with 3 allowed failures, so that also boils down to about 3 minutes.

I thought about submitting my health check as a PR, but until now I didn't as I wasn't sure what to make configurable. Depending on the URL, you want different timeouts. Or you want more checks. Or... So I was a bit worried about the amount of extra parameters :-)

Anyway: **a health check is a good idea**, but it is something that's potentially a bit difficult to get right for everyone.

ahennr:
Thanks for your detailed feedback @reinout

First of all, I chose the parameters and the URL of the  health check for the default GeoServer docker image settings without changing any configurations or adding any further data sources.

Regarding your notes on the URL:
The URL should point to a resource that is available as soon as the GeoServer has started up correctly.
In my opinion, a small resource delivered via the GeoServer application (image resource or a simple SLD) is sufficient at this point, as this shows whether the GeoServer / Tomcat could be started correctly.
A check for the correct configuration of the data sources (e.g. database or external service) is rather a specific check of the concrete installation than of a general GeoServer instance. Beside, GeoServer will also start if the database is not available, but will mark the corresponding DataSource as inactive.
If you want to test an appropriately configured layer / data source, you could implement this by calling a layer of a [virtual service](https://docs.geoserver.org/stable/en/user/configuration/virtual-services.html#filtering-by-layer) (as a WMS GetMap with a corresponding layer) so that the data source is also queried.

As an alternative to using a static image, the REST API could be used. However, as far as I know, explicit permissions to access the API have to be set here (or usage of Basic auth in URL), which are not activated by default. Management endpoints (`/health`, `/info`), such as those offered by [Spring Boot Actuator](https://docs.spring.io/spring-boot/docs/2.5.6/reference/html/actuator.html#actuator.endpoints.enabling), would be helpful in this context. However, as far as I know, these are not implemented in GeoServer, even in a different way.

I totally agree with you on the timeout point and it makes sense to make them parameterizable as well. I would come up with a corresponding commit if requested.

Are there any other opinions on this?

buehner:
Thx for the discussion. I really like this and will merge it now.

@reinout It seems to be possible to easily overwrite the healtcheck values defined in the Dockerfile. This can either be done by using a healthcheck definition in the docker-compose file as added here: 9b80e8c

It tested this by changing the test in the docker-compose to `test: curl --fail "http://localhost:8080/geoserver/tiger/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=tiger%3Apoly_landmarks&maxFeatures=1&outputFormat=application%2Fjson" || exit 1`

after that you can see that the geosjon is used as healthcheck: `docker inspect abc | jq '.[].State.Health.Log'`

Or by passing/overwriting the healthcheck values directly in a `docker run` command: https://docs.docker.com/engine/reference/run/#healthcheck

reinout:
You're right. Customizing the health check is easy.
What I'm thinking: how much customization is needed? Even the `HEALTHCHECK_URL` isn't really needed, I think. **If** you want a different url, you probably also want a different timeout or frequency. Making all that configurable is a hassle.

My suggestion: just use the health check as-is, and remove the `HEALTHCHECK_URL` setting and don't add any more configurable settings. (And document how to customize the health check).

PR #40, opened 2023-12-13
Add trivy scan

hwbllmnn:
Adds an action that builds the image and performs a trivy security scan on it.


buehner:
Thank you. As this will not change the Dockerfile, but provides some background information. I will merge

PR #42, opened 2023-12-20
chore: do not fail with known vulns

buehner:
For the time being it should be fine to have the output of the github action (security scan) as a background info, but it should not be blocking in any way.


hwbllmnn (approved):

ahennr (approved):
👍🏼 

Maybe it makes sense to introduce a security policy [here](https://github.com/geoserver/docker/security), what do you think?

PR #45, opened 2024-02-07
Switch to official Tomcat 9 - JDK 11 base image 

ahennr:
In this PR a change of the base image to the official Tomcat 9 (`tomcat:9.0.85-jre11-temurin-jammy`) image is suggested (internally based on ubuntu 22.04). In particular, I see the following advantages of this change:

* By using this image, Java and Tomcat do not have to be installed manually and they are pre-configured in a consistent way
* Since the Tomcat image is regularly updated on the operating system level as well as on the Tomcat side, this can be used to continuously update the built images and release them with the same tag but different digest when updating the base image, if the GeoServer version remains the same.
  *  See also the discussion in https://osgeo-org.atlassian.net/browse/GEOS-11231 
* The resulting GeoServer image is about 38% smaller than the one currently provided


buehner:
Thank you. I like this change and will take some time soon to test this extensively to ensure that all available features are still working with such a "basic" change.

Dockerfile
```diff
@@ -1,21 +1,48 @@
-FROM ubuntu:22.04 as tomcat
+FROM tomcat:9.0.85-jre11-temurin-jammy
```
buehner, line 1: We could think about introducing build args like

```
ARG TOMCAT_VERSION=9.0.85
ARG JAVA_VERSION=11

FROM tomcat:${TOMCAT_VERSION}-jre{JAVA_VERSION}-temurin-jammy
```

This would give the build more flexibility.

We were previously using the `jdk` instead of `jre`. I don't know if there was a reason for this? Maybe @jodygarnett knows?

jodygarnett: I do not know, GeoServer is supposed to only require JRE.

aaime: Agreed. But it might be tricky to get the JRE going, as it often requires less dependencies, which end up tripping either the startup, the map rendering, or specifically, the point rendering (fondconfig/harfbuzz not being installed or not completely working). Try it out with "topp:states", if that works, it _should_ be fine.

buehner:
Would be great if you could resolve the conflicts @ahennr :pray:

Dockerfile
ahennr: The idea behind using the `JRE` image was that this would result in a smaller GeoServer image. I didn't realize that this could lead to side-effects in rendering for example. Thanks @aaime for this hint!

In 2de74a9bd1fff67b8c5d2a030938b7255b4536b1 I switched back to `JDK` in order to be consistent with the previous version and to avoid the difficulties of using the `JRE` (see above comment). Furthermore, due to the changes in https://github.com/geoserver/docker/pull/39, the `catalina.jar` has to be packed/unpacked using `jar` to exchange the version info which is not available in the `JRE` image.

ahennr:
> Would be great if you could resolve the conflicts @ahennr 🙏

🏁

Dockerfile
```diff
@@ -1,21 +1,48 @@
-FROM ubuntu:22.04 as tomcat
+FROM tomcat:9.0.85-jdk11-temurin-jammy
```
mprins, line 1:
```suggestion
FROM tomcat:9.0.86-jdk11-temurin-jammy
```

ahennr: Unfortunately, `tomcat:9.0.86` is not yet available as an official docker image. So I switched to `tomcat:9.0.85-jdk11-temurin-jammy` again.

As soon as a dependency update such as renovate (https://github.com/geoserver/docker/pull/48) is properly configured, this should be updated automatically / proposed as a PR.

Dockerfile
```diff
@@ -1,22 +1,48 @@
-FROM ubuntu:22.04 as tomcat
+FROM tomcat:9.0.91-jdk11-temurin-jammy
```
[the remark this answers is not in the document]
ahennr: According to [the discussion before](https://github.com/geoserver/docker/pull/45#discussion_r1484051698) I prefer to use `jdk` (especially comment https://github.com/geoserver/docker/pull/45#discussion_r1482500859). In addition, without JDK, it's not possible  (AFAIK) to alter the `server.info` string in the `catalina.jar`

README.md
```diff
@@ -1,17 +1,18 @@
...
-* GeoServer
-  * Support of custom fonts (e.g. for SLD styling)
-  * CORS support
-  * Support extensions
-  * Support additional libraries
-  * Support for PostgreSQL JNDI 
-  * Support for HTTPS
+* Based on the official [`tomcat` docker image](https://hub.docker.com/_/tomcat), in particular:
+  * Tomcat 9
+  * JRE11 (eclipse temurin)
```
buehner, line 7:
```suggestion
  * JDK11 (eclipse temurin)
```

buehner:
Thx a lot for the work @ahennr 

I really like this for different reasons:

* ~30% reduced image size (even with JDK)
* better support with https://github.com/geoserver/docker/pull/48
* simpler `Dockerfile`
* probably more...

I think I'll merge this soon (probably other PRs that might be merged before lead to new merge conflicts)

buehner:
I made some final extensive testing and everything was fine, so i will merge this now. Thx again :pray:

PR #46, opened 2024-02-07
fix: use sarif format in trivy results

ahennr:
Fixes the output format of `trivy` workflow.


buehner:
Thx, I will merge immediately

PR #47, opened 2024-02-08
Support tomcat root

csbarnet:


buehner:
Hi. Thank you for the PR. Could you add a small section to the README to explain this new feature?

Dockerfile
```diff
@@ -129,4 +130,4 @@ ENTRYPOINT ["/opt/startup.sh"]
 WORKDIR /opt
 
 HEALTHCHECK --interval=1m --timeout=20s --retries=3 \
-  CMD curl --fail $HEALTHCHECK_URL || exit 1
+  CMD if [ $WEBAPP_CONTEXT == "ROOT" ];then (curl --fail ROOT_HEALTHCHECK_URL || exit 1); else (curl --fail HEALTHCHECK_URL || exit 1);fi
```
reinout, line 133: You're missing the `$` signs for the variables:

```suggestion
  CMD if [ $WEBAPP_CONTEXT == "ROOT" ];then (curl --fail $ROOT_HEALTHCHECK_URL || exit 1); else (curl --fail $HEALTHCHECK_URL || exit 1);fi
```

Dockerfile
```diff
@@ -95,20 +96,20 @@ ENV ROOT_WEBAPP_REDIRECT=false
 ENV POSTGRES_JNDI_ENABLED=false
 ENV CONFIG_DIR=/opt/config
 ENV CONFIG_OVERRIDES_DIR=/opt/config_overrides
-ENV HEALTHCHECK_URL=http://localhost:8080/geoserver/web/wicket/resource/org.geoserver.web.GeoServerBasePage/img/logo.png
-
+ENV HEALTHCHECK_URL=http://localhost:8080/$WEBAPP_CONTEXT/web/wicket/resource/org.geoserver.web.GeoServerBasePage/img/logo.png
+ENV ROOT_HEALTHCHECK_URL=http://localhost:8080/web/wicket/resource/org.geoserver.web.GeoServerBasePage/img/logo.png
```
reinout, line 100: In most files, you just have to swap `geoserver` for `$WEBAPP_CONTEXT`. Nice.

Only for the healthcheck, this extra `ROOT_HEALTHCHECK_URL` is needed in combination with an if/else in the actual health check, complicating it somewhat. I'd personally like it more if the extra env var and if/else would not be needed. I can't find a quick way to do that, though :-)

One possible idea: set `WEBAPP_CONTEXT` to an empty value if you want it in the root? For the "mv/mkdir" stuff, a `mkdir /some/dir/${WEBAPP_CONTEXT:-ROOT}/somewhere` would then be enough. Just brainstorming here.

csbarnet:
@reinout thank you for the comments and suggestions. 

While my proposed change (with a couple of fixes) works ok for building a custom image, but I'd like to try another approach that allows someone to change the context path after the image is built. I think that I can do this by either: turning off autodeploy and deployonstartup in server.xml and changing the context.xml or by moving the geoserver webapp files out of the /webapp path and altering the context.xml (according to tomcat 9 docs). 

Is one of these approaches preferred? Also, should I close this pull request and start fresh, or just make those changes to this pull request branch?

csbarnet:
This version keeps all of the geoserver files in /webapps/geoserver, but changes the context path by setting the docBase and path attributes in a context element in server.xml. 

AFAIK, this is the only way to accomplish this without moving the docBase for the app context out of the host's appBase.

From the Tomcat docs:

> If you want to deploy a WAR file or a directory using a context path that is not related to the base file name then one of the following options must be used to prevent double-deployment:
> 
> Disable autoDeploy and deployOnStartup and define all Contexts in server.xml
> Locate the WAR and/or directory outside of the Host's appBase and use a context.xml file with a docBase attribute to define it.

Some relevant Tomcat 9 documentation is here:
[](https://tomcat.apache.org/tomcat-9.0-doc/config/context.html#Naming)
[](https://tomcat.apache.org/tomcat-9.0-doc/config/context.html#Defining_a_context)

Dockerfile
```diff
@@ -8,6 +8,7 @@ ARG CORS_ALLOWED_HEADERS=*
 ARG CORS_ALLOW_CREDENTIALS=false
 
 # Environment variables
+ENV TOMCAT_VERSION=$TOMCAT_VERSION
```
csbarnet, line 11: one of the commands later in the Dockerfile uses TOMCAT_VERSION, but the ARG value doesn't propagate across multistage builds.

buehner: Thx. We added this line here: https://github.com/geoserver/docker/pull/51

spike83:
works like charm on my k8s as well.
A healthcheck which can be filtered or does not appear in the monitor would be nice. I think the former with the image is a bit light and this one is a bit heavy (and appears in the monitor). Not sure if I missed it but have not seen any healthcheck implemented into geoserver. I think this would be great and could maybe also include datasources.

buehner:
Seems we have merge conflicts now, so this has to be resolved

csbarnet:
@buehner, I finally got around to fixing the merge conflicts. I'm currently testing.

buehner:
@csbarnet Thx a lot. I also tested this again and it works fine for me. I am ready to merge this if your testing was also fine? Just let me know

csbarnet:
@buehner My testing was also fine. thank you!

PR #51, opened 2024-02-20
fix: set tomcat version as environment variable

buehner:
This is a possible fix for the build issues mentioned in https://github.com/geoserver/docker/pull/39#issuecomment-1953750070 (and https://github.com/geoserver/docker/pull/47#discussion_r1488462942)


ahennr:
👍🏼

PR #56, opened 2024-05-14
fix: allow unicode filenames

hwbllmnn:
GeoServer won't start if you have e.g. a style with an umlaut in the filename. With this PR this is fixed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ ENV CATALINA_OPTS="\$EXTRA_JAVA_OPTS \
 # init
 RUN apt update \
     && apt -y upgrade \
-    && apt install -y --no-install-recommends openssl unzip gdal-bin wget curl openjdk-11-jdk gettext \
+    && apt install -y --no-install-recommends locales openssl unzip gdal-bin wget curl openjdk-11-jdk gettext \
     && apt clean \
     && rm -rf /var/cache/apt/* \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


buehner:
Is there a reason to not add this to a [previous stage where apt is already used](https://github.com/geoserver/docker/pull/56/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R33-R38)?

If not: Could you give it a try by applying the change there?

hwbllmnn:
> Thank you! Is there a reason to not add this to the [previous stage where apt is already used](https://github.com/geoserver/docker/pull/56/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R33-R38)?
> 
> If not: Could you give it a try by applying the change there?

Yes, the change needs to be available at runtime, so adding the package to a previous stage doesn't help.

dnlkoch (approved):

hwbllmnn:
> > Thank you! Is there a reason to not add this to the [previous stage where apt is already used](https://github.com/geoserver/docker/pull/56/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R33-R38)?
> > If not: Could you give it a try by applying the change there?
> 
> Yes, the change needs to be available at runtime, so adding the package to a previous stage doesn't help.

Alright, I didn't realize that the actual image starts from the first stage, I moved the apt stuff there like you suggested.

PR #57, opened 2024-05-24
Set admin credentials with environment variables

dmenneck:
This project lacked the option of specifying the username and password of the admin user via environment variables. With this PR it's possible to set the admin credentials like so:

```sh
podman run -d -p 8081:8080 -e GEOSERVER_ADMIN_USER=admin -e GEOSERVER_ADMIN_PASSWORD=geoserver <image>
```
The changes are heavily influenced by preliminary work of [meggsimum](https://github.com/meggsimum/geoserver-docker/) and [kartoza](https://github.com/kartoza/docker-geoserver).

PR #61, opened 2024-07-15
fix: use more slashes to make script more robust

simonseyock:
As `//` inside of a path will be ignored in the shell, it might be safer to use more slashes.

Don't know if this is desired, it makes the whole thing a little more robust imho.

We had a problem where the security directory was copied over everytime and the masterpasswd changed on every restart.

### DIFF
--- a/update_credentials.sh
+++ b/update_credentials.sh
@@ -11,16 +11,16 @@ fi;
 
 # copy over default security folder to data dir (if not existing)
 if [ ! -d "${GEOSERVER_DATA_DIR}/security" ]; then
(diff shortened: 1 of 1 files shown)


buehner:
First thing i realize is: `APP_PATH_PREFIX` is not needed in my eyes. This was introduced in #57 and I did not see this when reviewing :sweat_smile: 

The feature was ported from two other geoserver docker projects and `APP_PATH_PREFIX` does not play any role within this project (like in the [the startup script of the origin project](https://github.com/search?q=repo%3Ameggsimum%2Fgeoserver-docker%20APP_PATH_PREFIX&type=code))

So I would suggest to fully remove all appearances of `APP_PATH_PREFIX`. Could you add that to your PR? @simonseyock

buehner:
Great, thx a lot. I tested your branch successfully in different scenarios and will merge now.

PR #6, opened 2022-08-14
Troubleshoot Dockerfile handling of data directory and extensions

jodygarnett:
In writing docs and updating readme I found a few things to fix:

* [x] installing extensions on startup now works (had incorrect download location)
* [x] global.xml required file check, install release data directory if needed (since it is already downloaded)
* [x] docker-compose-demo referencing environmental variables no longer used
* [x] some readme instructions out of date or untested

PR #62, opened 2024-07-17
feat: set GEOSERVER_REQUIRE_FILE to currently used $GEOSERVER_DATA_DIR in startup.sh

simonseyock:
There exists a problem where the default settings are copied over the existing geoserver data dir if a custom data dir is used.

If only GEOSERVER_DATA_DIR and neither SKIP_DEMO_DATA nor GEOSERVER_REQUIRE_FILE are set to any values. The GEOSERVER_REQUIRE_FILE will still be set to the default location because it is set in the Dockerfile and not in the startup script.

I assume that GEOSERVER_REQUIRE_FILE is not used outside the startup script is that correct?

The proposed solution here is one way to solve the issue without dropping the environment variable, but honestly as it is only used for checking if the demo data should be copied, I would suggest to drop the variable all together and rather do something like this:

```bash
## install release data directory if needed before starting tomcat
if [ "${SKIP_DEMO_DATA}" = "false" ] && [ ! -f "$GEOSERVER_DATA_DIR/global.xml" ]; then
  echo "Initialize $GEOSERVER_DATA_DIR from data directory included in geoserver.war"
  cp -r $CATALINA_HOME/webapps/geoserver/data/* $GEOSERVER_DATA_DIR
fi
```

instead of

```bash
## Skip demo data
if [ "${SKIP_DEMO_DATA}" = "true" ]; then
  unset GEOSERVER_REQUIRE_FILE
elif [ -z "$GEOSERVER_REQUIRE_FILE" ]; then
  export GEOSERVER_REQUIRE_FILE="$GEOSERVER_DATA_DIR/global.xml"
fi

## install release data directory if needed before starting tomcat
if [ ! -z "$GEOSERVER_REQUIRE_FILE" ] && [ ! -f "$GEOSERVER_REQUIRE_FILE" ]; then
  echo "Initialize $GEOSERVER_DATA_DIR from data directory included in geoserver.war"
  cp -r $CATALINA_HOME/webapps/geoserver/data/* $GEOSERVER_DATA_DIR
fi
```


simonseyock:
Hint: I do not see GEOSERVER_REQUIRE_FILE in here: https://docs.geoserver.org/stable/en/user/configuration/properties/index.html

buehner:
> I assume that GEOSERVER_REQUIRE_FILE is not used outside the startup script is that correct?

Yes. I am pretty sure that this env only exists in this docker project, but not in main project or somewhere else.

buehner:
`GEOSERVER_REQUIRE_FILE` was first introduced in https://github.com/geoserver/docker/pull/6 and later `SKIP_DEMO_DATA` in https://github.com/geoserver/docker/pull/9

Maybe this helps to understand the scenario.

buehner:
> The proposed solution here is one way to solve the issue without dropping the environment variable, but honestly as it is only used for checking if the demo data should be copied, I would suggest to drop the variable all together and rather do something like this:
> 
> ```shell
> ## install release data directory if needed before starting tomcat
> if [ "${SKIP_DEMO_DATA}" = "false" ] && [ ! -f "$GEOSERVER_DATA_DIR/global.xml" ]; then
>   echo "Initialize $GEOSERVER_DATA_DIR from data directory included in geoserver.war"
>   cp -r $CATALINA_HOME/webapps/geoserver/data/* $GEOSERVER_DATA_DIR
> fi
> ```

I agree. Seems like `SKIP_DEMO_DATA` is the only flag we should use to control whether the demo data should be installed or not.

For now, I'd like to merge this MR as it fixes the problem. We could target the proposed solution to fully drop the `GEOSERVER_REQUIRE_FILE` in a follow up PR. Fine for you @simonseyock ?

jodygarnett:
GEOSERVER_REQUIRE_FILE was specifically introduced to mitigate some kind of security vulnerability with Docker images being started up against an empty GEOSERVER_DATA_DIR. It was reported before we did CVEs and I did not use Docker at the time so I do not remember the *exact* attack.

But this is used to confirm the location has a global.xml or some other file and thus we are *sure* it contains a data directory.  The use is optional because some people start up GeoServer with an empty data directory and then use the REST API to configure. 

The workflow we have, where the startup script unpacks the default data directory, so we are sure there is a data directory in place by the time GeoServer runs, should be compatible with and safe when combined with GEOSERVER_REQUIRE_FILE

jodygarnett:
@simonseyock for this statement:

> I assume that GEOSERVER_REQUIRE_FILE is not used outside the startup script is that correct?

Is my answer sufficient? The setting should always be provided and indicates a file GeoServer checks when it starts up. If the file is not present GeoServer will refuse to start up to avoid writing files into "the wrong" location.

simonseyock:
@jodygarnett sorry for not responding for so long. I updated the PR with your suggestion.

The only problem I see is if people have set this to full path, this would not work anymore. I.e. this is a breaking change.

jodygarnett:
I am sorry @simonseyock I only entered the chat to say that GEOSERVER_REQUIRED_FILE is part of GeoServer. I do not understand exactly what the subject of this PR discussion is. 

Q: Given that the location of GEOSERVER_REQUIRED_FILE is mapped to a consistent location in the image filesystem can the script provide this value?

For the workflow of starting up against an empty directory there are options:

a) copy the demo data directory into place so there is a starting point. This is compatible with GEOSERVER_REQUIRED_FILE mapped to $GEOSERVER_DATA_DIR/global.xml

b) do not copy the demo data directory, the user is supplying their own. This is also compatible with GEOSERVER_REQUIRED_FILE mapped to $GEOSERVER_DATA_DIR/global.xml

c) do not copy the demo data directory, the user wants the geoserver defaults (which honestly are not very good). For this case they need to set GEOSERVER_REQUIRED_FILE to empty to allow GeoServer to startup. This is the only case where GEOSERVER_REQUIRE_FILE is not desired.

Q: Can we identify between options (b) and (c) above?

simonseyock:
> Q: Given that the location of `GEOSERVER_REQUIRE_FILE` is mapped to a consistent location in the image filesystem can the script provide this value?

The problem is that the `GEOSERVER_DATA_DIR` can be changed via environment variables and the `GEOSERVER_REQUIRE_FILE` then points to the wrong location because it is set in the Dockerfile to a default value and is not updated.

To differentiate between option (b) and (c) there exists the `SKIP_DEMO_DATA` environment variable.

simonseyock:
1. Do not initialize `GEOSERVER_REQUIRE_FILE` in the Dockerfile. It can then be initialized to `$GEOSERVER_DATA_DIR/global.xml` in the startup script.
2. This means the user cannot unset the `GEOSERVER_REQUIRE_FILE`. If a user really wants the default behaviour (starting with an empty data dir), we could introduce an environment variable `USE_REQUIRE_FILE`.

So in the Dockerfile we would have:
```
ENV GEOSERVER_REQUIRE_FILE=
ENV USE_REQUIRE_FILE=true
ENV GEOSERVER_DATA_DIR=/opt/geoserver_data/
ENV SKIP_DEMO_DATA=false
```

Assuming the user wants to change the `GEOSERVER_DATA_DIR`, he can achieve the options like this:
(a) Do not set any environment variable
(b) Set `SKIP_DEMO_DATA` to true.
(c) Set `SKIP_DEMO_DATA` to true and set `USE_REQUIRE_FILE` to false

simonseyock:
I talked with @buehner and pushed changes with a new env variable SET_GEOSERVER_REQUIRE_FILE.

This is how it compares to the old beahvior:

old behavior:

* SKIP_DEMO_DATA=true -> GEOSERVER_REQUIRE_FILE will be ignored, even if set to a value - That is not really relevant because the require file will always exist as it was copied there.
* GEOSERVER_DATA_DIR=/somwehere/else -> default GEOSERVER_REQUIRE_FILE will be set to wrong value

new behavior:

* GEOSERVER_REQUIRE_FILE -> If this value is set, it will always be used, even if SKIP_DEMO_DATA or SET_GEOSERVER_REQUIRE_FILE are set. By default is it not set
* SET_GEOSERVER_REQUIRE_FILE -> If this value is set to true and SET_GEOSERVER_REQUIRE_FILE is not set, it will be set to the `global.xml` in the GEOSERVER_DATA_DIR
* SKIP_DEMO_DATA=true and SET_GEOSERVER_REQUIRE_FILE=true -> GEOSERVER_REQUIRE_FILE will be set to the global.xml of the demo data
* SKIP_DEMO_DATA=true and SET_GEOSERVER_REQUIRE_FILE=false -> old behavior of SKIP_DEMO_DATA=true

What I like about this change is, that it does not affect existing solutions

startup.sh
```diff
@@ -48,8 +43,17 @@ DEFAULT_HEALTHCHECK_URL="http://${DEFAULT_HEALTHCHECK_URL}"
 # write the healthcheck URL to a file that geoserver user has access to but is not served by tomcat
 echo "${HEALTHCHECK_URL:-$DEFAULT_HEALTHCHECK_URL}" > $CATALINA_HOME/conf/healthcheck_url.txt
 
+# set geoserver require file to global.xml
+if [ "SET_GEOSERVER_REQUIRE_FILE" = true ]; then
```
buehner, line 47: Please fix this to `$SET_GEOSERVER_REQUIRE_FILE`
Currently no ENV is used here :-)

simonseyock: oups :D

buehner:
Thanks a lot @simonseyock 

I tested a lot and it looks good to me, so this should not affect existing solutions -> I will merge now

PR #85, opened 2024-09-11
Build and release Geoserver multi-platform for ARM64 and AMD64.

alfonx:
Currently, the docker images at https://docker.osgeo.org are only compatible with AMD64 platform. Basically, they run on Intel and AMD, not on ARM. 

ARM CPUs are winning in popularity, and hosters like Hetzner offer them as root-servers, too.

This pull request will change the release.sh to build and push images for both platforms in parallel. After applying this PR, the release-Script will fail unless the docker-daemon has been set up for multi-platform build, as described here 
https://docs.docker.com/build/building/multi-platform/#qemu

tl/tr: 
 * Add to /etc/docker/daemon.json `{"features": {"containerd-snapshotter": true }}` and restart docker, e.g. `systemctl restart docker`
* Create a custom builder:
 `docker buildx create --name container-builder --driver docker-container --use --bootstrap`

After that, running build/release.sh should build and push images for both platforms. I cannot test the push, since I have not Nexus Repository to test! Building does not take much longer.

Whoever is releasing new geoserver docker images regularly, will need a multi-platform docker setup.

NB: With our docker-repository provider quay.io, we experienced, that the first push to a repository had to be multi-platform, for the repository to support multi-platform! I think we simply have to try with Nexus here.

Of course I am happy to test on ARM as soon and images are pushed.

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -85,7 +85,7 @@ if [[ $1 == *build* ]]; then
         -t $TAG .
     fi
   else
-    echo "docker build ---platform linux/amd64,linux/arm64 -build-arg GS_VERSION=$VERSION --build-arg GS_BUILD=$BUILD -t $TAG ."
+    echo "docker build --platform linux/amd64,linux/arm64 --build-arg GS_VERSION=$VERSION --build-arg GS_BUILD=$BUILD -t $TAG ."
     docker build \
       --platform linux/amd64,linux/arm64 \
       --build-arg GS_VERSION=$VERSION \
--- a/build/release.sh
+++ b/build/release.sh
@@ -49,6 +49,9 @@ else
   fi
 fi
 
+# Prerequisite for Multi-Arch via QEMU:
+docker run --privileged --rm tonistiigi/binfmt --install all
+
 echo "Release from branch $BRANCH GeoServer $VERSION as $TAG"
 
 # Go up one level to the Dockerfile
@@ -60,9 +63,10 @@ if [[ $1 == *build* ]]; then
     echo "  nightly build from https://build.geoserver.org/geoserver/$BRANCH"
     echo
     if [[ "$BRANCH" == "main" ]]; then
-      echo "docker build --build-arg GS_VERSION=$VERSION --build-arg GS_BUILD=$BUILD -t $TAG ."
+      echo "docker build --platform linux/amd64,linux/arm64 --build-arg GS_VERSION=$VERSION --build-arg GS_BUILD=$BUILD -t $TAG ."
       # todo: --no-cache-filter download,install
       docker build \
+        --platform linux/amd64,linux/arm64 \
         --build-arg WAR_ZIP_URL=https://build.geoserver.org/geoserver/main/geoserver-main-latest-war.zip \
         --build-arg STABLE_PLUGIN_URL=https://build.geoserver.org/geoserver/main/ext-latest \
         --build-arg COMMUNITY_PLUGIN_URL=https://build.geoserver.org/geoserver/main/community-latest \
@@ -70,8 +74,9 @@ if [[ $1 == *build* ]]; then
         --build-arg GS_BUILD=$BUILD \
         -t $TAG .
     else
-      echo "docker build --build-arg GS_VERSION=$VERSION --build-arg GS_BUILD=$BUILD -t $TAG ."
+      echo "docker build --platform linux/amd64,linux/arm64 --build-arg GS_VERSION=$VERSION --build-arg GS_BUILD=$BUILD -t $TAG ."
       docker build \
+        --platform linux/amd64,linux/arm64 \
         --build-arg WAR_ZIP_URL=https://build.geoserver.org/geoserver/$BRANCH/geoserver-$BRANCH-latest-war.zip \
         --build-arg STABLE_PLUGIN_URL=https://build.geoserver.org/geoserver/$BRANCH/ext-latest \
         --build-arg COMMUNITY_PLUGIN_URL=https://build.geoserver.org/geoserver/$BRANCH/community-latest \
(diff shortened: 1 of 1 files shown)


buehner:
Thank you for the PR. I really like the idea to extend the support to other platforms. I am only partially familiar with the build environment (Jenkins on build.geoserver.org), so I think we need support from others (with more rights to do such high level configuration). @jodygarnett could you help here regarding the QEMU steps described above?

jodygarnett:
@tbarsballe do you have access to ``/etc/docker/daemon.json`` or do we contact GeoSolutions system admin...

tbarsballe:
I do have access, but the command
```
docker buildx create --name container-builder --driver docker-container --use --bootstrap
```
Is failing. Not sure if it's a version issue (we're on Docker 24.0.5), or some missing prerequisite.

tbarsballe:
Though, reading through the linked documentation, it sounds like we need to to [enable the containerd image store](https://docs.docker.com/build/building/multi-platform/#enable-the-containerd-image-store) (which I have done) **OR**  [create a custom builder](https://docs.docker.com/build/building/multi-platform/#create-a-custom-builder), not both, so we might be good to go.

alfonx:
Thanks @tbarsballe .. I think a test-run could look like this:

 * Change the Jenkins-Job to do "./release.sh build" instead of "./release.sh buildandpublish "
 * Merge the PR
 * Check that it builds in the jenkins job
 * Get hold of someone with r/w access to https://docker.osgeo.org/ so that we could delete/correct any broken pushs
 * Try to run  "./release.sh buildandpublish" any test pulling and starting on ARM64 + AMD64

jodygarnett:
I am checking in with this activity for 2.26.0 release. I am only releasing locally as something is troubling Jenkins (see geoserver-devel email list).

alfonx:
>Could I merge this and do a manual release with ARM64?

Yes.

buehner:
It seems that the renovate bots of geoserver users are suggesting to update to geoserver 2.26.0 as the docker image is already published: [docker pull docker.osgeo.org/geoserver:2.26.0`](https://docker.osgeo.org/#browse/browse:geoserver-docker:v2%2Fgeoserver%2Ftags%2F2.26.0)

I tried `docker pull docker.osgeo.org/geoserver:2.26.0` and then i had a look at `docker image inspect docker.osgeo.org/geoserver:2.26.0` and realized: `"Architecture": "arm64" (where 2.25.3 still has `amd64`)

I guess this is somehow related to this MR? @jodygarnett @tbarsballe 

I am just asking what exactly happened? Is it a different config on jenkins? Or the result of a manual action?

jodygarnett:
Enabling this option to try out locally:

image

tbarsballe:
Independant from the above issue, it appears that enabling the containerd store on Jenkins broke the existing builds ([for example](https://build.geoserver.org/view/release/job/geoserver-release-docker/1277/console)) as follows:
```
parent snapshot sha256:334047a74abd6cc7c7191ab195b9a909f4ff24f50ee6f82cb9780fc8ca8a82a5 does not exist: not found
```
(The actual snapshot sha changes each build, so it appears to be an intermediate snapshot)

While this error makes some amount of sense as containerd images are incompatible with standard images, removing all stored images from the server was not able to resolve this failure, so something else seems to be going on.

Reverting the change to daemon.json and disabling the containerd store fixed the build failure, so without a resolution to the above issue I don't think we can move forward with this change.

petersmythe:
Is there any progress on this?  I too use arm64 and currently need to build the image myself, so I would appreciate this as well.

jodygarnett:
I am not sure @petersmythe I was also looking to see if there was something to do here.

maciejsk:
Is there any progress? very necessary thing. allows to reduce costs.

jodygarnett:
> Is there any progress? very necessary thing. allows to reduce costs.

Are you in position to resolve conflicts and test?

jodygarnett:
I think this needs an infrastructure change eh?

byander:
Some news about this PR? 
I tried using it on ARM and not working for version 2.28.X

alfonx:
Great news! Geoserver @-latest@ tags are running on ARM now!
I just checkd the manifests and did a test on my Ampere Altra Q80 CPU, works!

| Tag docker.osgeo.org/geoserver: | linux/amd64 | linux/arm64 |
|---|---|---|
| 3.0-latest | ✅ | ✅ |
| 2.28-latest | ✅ |  ✅ |
| 3.0.0 | ✅ | ✅ |
| 3.0-RC | ✅ | ✅ |
| 3.0-latest-gdal | ✅ | ✅ |
| 3.0.x | ✅ | ❌ |
| 2.28.4 | ✅ | ❌ |

So ARM is avaibale for Geoserver 3 and 2.28, but one has to take care which tag to use!
I will close this ticket. Thank you very much @petersmythe !

aaime:
I'd be really curious to see a performance comparison between arm and amd, cost and performance wise, with an actual benchmark/load test. Has anyone here done one?

petersmythe:
> Great news! Geoserver @-latest@ tags are running on ARM now! I just checkd the manifests and did a test on my Ampere Altra Q80 CPU, works!
> 
> Tag docker.osgeo.org/geoserver:	linux/amd64	linux/arm64
> 3.0-latest	✅	✅
> 2.28-latest	✅	✅
> 3.0.0	✅	✅
> 3.0-RC	✅	✅
> 3.0-latest-gdal	✅	✅
> 3.0.x	✅	❌
> 2.28.4	✅	❌
> So ARM is avaibale for Geoserver 3 and 2.28, but one has to take care which tag to use! I will close this ticket. Thank you very much @petersmythe !

Thanks for pointing out that 2.28.4 was not built for arm64, that mistake has been corrected now.

Also, are you certain that 3.0.x is not built for arm64, I can see that it is:

>  docker manifest inspect docker.osgeo.org/geoserver:3.0.x | grep arm64
>             "architecture": "arm64",

PR #94, opened 2024-09-23
chore: use latest gs everywhere

buehner:
Just updating to latest GeoServer as default. Also updates doc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ARG CORS_ALLOW_CREDENTIALS=false
 ARG CORS_ENABLED=false
 ARG GS_BUILD=release
 ARG GS_DATA_PATH=./geoserver_data/
-ARG GS_VERSION=2.25.2
+ARG GS_VERSION=2.25.3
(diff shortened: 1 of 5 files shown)


reinout:
Shouldn't this be `2.26.0`?

buehner:
> Shouldn't this be `2.26.0`?

As 2.26.0 is not yet released, the documented steps would not work currently (docker pull etc). But once 2.26.0 is released, we should update again. Would be great to have some kind of automatization for this in future.

In my eyes theses changes are more "pro forma" as you can always use the build argument for the gs version you like to build. But of course it makes sense to have it up-to-date.

reinout:
2.26.0 isn't released yet? Why is there a (failing, but that's known) 2.26.0 geoserver docker release then? I assumed that a docker release would only be made from a released geoserver version.

If my assumption is wrong, I have to change my own renovatebot setup, as it happily picked up the 2.26.0. But if the geoserver docker release is made before there's an actual release: that's not what was done up till now (as far as I could see).

buehner:
Ahh, interesting. I did not know that. Just had a look at https://geoserver.org/download/

So I guess the release process i currently ongoing...

PR #97, opened 2024-09-24
Allow running tomcat as non-root

jeanpommier:
It is a good practice to allow running a container as unprivileged user. 

To preserve backward compatibility and provide flexibility, this feature is added at the entrypoint level and is set using environment variables.
- RUN_UNPRIVILEGED=true activates the unprivileged mode with default uid:gid as 999:999
- RUN_WITH_USER_UID allows to set the uid used for tomcat user
- RUN_WITH_USER_UID allows to set the gid used for tomcat group
- CHANGE_OWNERSHIP_ON_FOLDERS accepts a space-separated list of folder on which a chmod will be run, changing (recursively) the ownership for the tomcat user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,10 @@ ENV HEALTHCHECK_URL=''
 ENV INSTALL_EXTENSIONS=false
 ENV POSTGRES_JNDI_ENABLED=false
 ENV ROOT_WEBAPP_REDIRECT=false
+ENV RUN_UNPRIVILEGED=false
+ENV RUN_WITH_USER_UID=
+ENV RUN_WITH_USER_GID=
+ENV CHANGE_OWNERSHIP_ON_FOLDERS="/opt $GEOSERVER_DATA_DIR"
 ENV SKIP_DEMO_DATA=false
 ENV STABLE_EXTENSIONS=''
 ENV STABLE_PLUGIN_URL=$STABLE_PLUGIN_URL
@@ -62,7 +66,7 @@ WORKDIR /tmp
 RUN set -eux \
     && export DEBIAN_FRONTEND=noninteractive \
     && apt-get update \
-    && apt-get install -y --no-install-recommends openssl unzip curl locales gettext \
+    && apt-get install -y --no-install-recommends openssl unzip curl locales gettext gosu \
     && apt-get clean \
     && rm -rf /var/cache/apt/* \
     && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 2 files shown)


jeanpommier:
Commented in the Dockerfile is an alternative solution where it is set during the build. This is more standard but less flexible. I'll remove it if we opt for the currently proposed solution, but I kept it for informational purpose.

buehner:
Thank you very much. This is a great feature that is requested by many users. I really like (and support) the approach to provide a backward compatible solution for this. I checked out your branch and played around locally.

**Backward compatibility (old)**

When trying to use it backward compatible ("as before", i.e. without using any of the new env variables), I currently end up (when starting the image) with: `error: failed switching to "tomcat": unable to find user tomcat: no matching entries in passwd file`

I guess the reason is that `exec gosu tomcat ...` is always using the `tomcat` user, which does not exist in the default/old scenario. (It works when changing to `exec gosu root ...`)

I think this should be fixed by checking for the current scenario and running as either `root` or `tomcat`. Would be great if you could add such a fix @jeanpommier 

**Running unprivileged (new)**

This seems to work fine. I tested by using only `RUN_UNPRIVILEGED=true`, but also by setting UID and GID explicitly. Could not find any issues.


@jodygarnett Do you need JIRA tickets to track such changes for future release notes (as mentioned in https://github.com/geoserver/docker/pull/39#issuecomment-2372483576 )?

jeanpommier:
> I guess the reason is that `exec gosu tomcat ...` is always using the `tomcat` user, which does not exist in the default/old scenario. (It works when changing to `exec gosu root ...`)

Oh, my bad, that was a late addition, I forgot to test it. Fixing it right now

jeanpommier:
Thanks for your support @buehner !
I realized I didn't do anything documentation-wise. We probably should at least document the env vars in this project's readme (should I do another PR for this ?). And ideally add something in the [geoserver doc](https://docs.geoserver.org/latest/en/user/installation/docker.html) ?

buehner:
Yes. You are right. Would be great if you could add this in a follow up PR for this repo at least (in the README?). Thx :pray:

jeanpommier:
Hi @buehner 

done by https://github.com/geoserver/docker/pull/106

pazepaze:
In case someone else runs into this: Note that setting the user in the entry point script is not sufficient to run in a kubernetes cluster that uses the `restricted` Pod Security Standard: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
The user would need to be set in the Dockerfile.

PR #102, opened 2024-10-01
GEOS-11555: fix installation of extensions in various configurations

spike83:
With this PR, I want to address the issues [GEOS-11555](https://osgeo-org.atlassian.net/browse/GEOS-11555) and [GEOS-11258](https://osgeo-org.atlassian.net/browse/GEOS-11258) about some inabilities of the extension installer script.

In my opinion, it should be possible to add community modules also to official images. So, I added a warning message instead of blocking it.

Also, I tried to get rid of the confusion in the naming of the variable `INSTALL_EXTENSIONS`  which was controlling the download behavior. Thus, I introduced a variable `DOWNLOAD_EXTENSIONS` which controls this and made the `INSTALL_EXTENSIONS` to really control the installation of extensions.

On one point, I'm not sure: Should we fail the startup when we are unable to download extensions?
On one hand, the full functionality might not he available and cannot be detected with any sort of health check and on the other hand, maybe it's better to run without an extension than not run at all.


README.md
```diff
@@ -226,25 +240,24 @@ Following is the list of the all the environment variables that can be passed do
...
+| DOWNLOAD_EXTENSIONS | Indicates whether additional GeoServer extensions should be installed | `false` |
+| INSTALL_EXTENSIONS | Indicates whether additional GeoServer extensions should be installed | `true` |
 | WAR_ZIP_URL | Specifies the URL for a GeoServer Web Archive (WAR) file | |
 | STABLE_EXTENSIONS | Specifies stable GeoServer extensions | |
 | STABLE_PLUGIN_URL | Specifies the URL for downloading the latest stable GeoServer plugins | `https://build.geoserver.org/geoserver/2.25.x/ext-latest` |
 | COMMUNITY_EXTENSIONS | Specifies community-contributed GeoServer extensions | |
-| COMMUNITY_PLUGIN_URL | Specifies the URL for downloading the latest community-contributed GeoServer plugins | `https://build.geoserver.org/geoserver/2.25.x/community-latest` |
+| COMMUNITY_PLUGIN_BASE_URL | Specifies the **base** URL for downloading the latest community-contributed GeoServer plugins. | `https://build.geoserver.org/geoserver/` | |
+| COMMUNITY_PLUGIN_URL | Specifies the URL for downloading the latest community-contributed GeoServer plugins | `${COMMUNITY_PLUGIN_BASE_URL}2.25.x/community-latest` |
```
reinout, line 250: I think it is clearer to end both URLs in the same way (with or without slash). Without slash is visually clearer (and matches the original `COMMUNITY_PLUGIN_URL`).

```suggestion
| COMMUNITY_PLUGIN_BASE_URL | Specifies the **base** URL for downloading the latest community-contributed GeoServer plugins. | `https://build.geoserver.org/geoserver` | |
| COMMUNITY_PLUGIN_URL | Specifies the URL for downloading the latest community-contributed GeoServer plugins | `${COMMUNITY_PLUGIN_BASE_URL}/2.25.x/community-latest` |
```

README.md
```diff
@@ -226,25 +240,24 @@ Following is the list of the all the environment variables that can be passed do
...
 | SKIP_DEMO_DATA | Indicates whether to skip the installation of demo data provided by GeoServer | `false` |
 | ROOT_WEBAPP_REDIRECT | Indicates whether to issue a permanent redirect to the web interface | `false` |
 | HEALTHCHECK_URL | URL to the resource / endpoint used for `docker` health checks | `http://localhost:8080/geoserver/web/wicket/resource/org.geoserver.web.GeoServerBasePage/img/logo.png` |
 | GEOSERVER_ADMIN_USER | Admin username |   |
 | GEOSERVER_ADMIN_PASSWORD | Admin password |  |
-
-The following values cannot really be safely changed (as they are used to download extensions and community modules as the docker image first starts up).
-| VAR NAME | DESCRIPTION | SAMPLE VALUE |
-|--------------|-----------|------------|
+| WGET_OPTS | Options for the `wget` command | `--no-check-certificate` |
```
reinout, line 258: Wasn't it your intention to remove this `--no-check-certificate`? (Which is a good idea to remove in my opinion :-) )

README.md
reinout: Although... I now notice that `ADDITIONAL_LIBS_DIR` seems to include the trailing slash. In that case, end both of these URLs with a slash?

install-extensions.sh
```diff
@@ -4,51 +4,98 @@
...
+      echo "Installed all jar files from ${ADDITIONAL_LIB}"
+    elif [[ $ADDITIONAL_LIB == *.jar ]]; then
+      cp ${ADDITIONAL_LIB} ${GEOSERVER_LIB_DIR}
+      echo "Installed ${ADDITIONAL_LIB}"
+    else
+      echo "Skipping ${ADDITIONAL_LIB}: unknown file extension."
+    fi
+}
+
+# Download stable plugins only if DOWNLOAD_EXTENSIONS is true
```
reinout, line 39: The setting applies to both stable and community plugins, right?

```suggestion
# Download plugins only if DOWNLOAD_EXTENSIONS is true
```

reinout:
Sorry for the multiple comments, I should have done it properly as a single review. I thought I had only one quick suggestion :-)

Fixing up the if/else errors and the community plugin installs in the install script seems like a good idea.

One thing that worries me a bit are the extra environment variables. There are a lot already. Could your fix work without the extra environment variables?

- `DOWNLOAD_EXTENSIONS`: if you specify an extension, it should be fine to download it if missing. Why the extra check?
- `COMMUNITY_PLUGIN_BASE_URL`: a variable in the bash script is probably handy, but I don't think it needs to be configurable?

README.md
spike83: Yes, it is removed by this PR, but there is the variable `WGET_OPTS` in case you would for some reason need to set some options for wget. By default, this is empty, just mentioning a sample value here.

spike83:
`DOWNLOAD_EXTENSIONS` is basically the functionality from the former  `INSTALL_EXTENSIONS` variable to disable the download. This will prevent some errors when run in an environment where the geoserver has no internet access.

`COMMUNITY_PLUGIN_BASE_URL` allows changing the server for the build artifacts but keep the logic of the structure underneath it.

I BTW missed one ;-))

It could make sense to split them into multiple tables to get some clarity and better overview. Suggesting Geoserver, Tomcat, Extensions

jodygarnett:
I would be very unhappy with installing community modules into a release intended for production - the point of them is to get feedback on snapshot / nightly build releases to support development. If people want to use extensions in production - stop messing around and fund / complete them.

Really we are already pushing our luck distributing these as nightly builds as they have not been given to the project steering committee for review and publication yet.

BUILD.md
```diff
@@ -65,27 +65,27 @@ Override these arguments to make use of build.geoserver.org nightly releases:
...
-Here is a working example for building 2.25.x nightly build::
+Here is a working example for building 2.27.x nightly build::
 ```
 docker build --no-cache-filter download,install \
-  --build-arg WAR_ZIP_URL=https://build.geoserver.org/geoserver/2.25.x/geoserver-2.25.x-latest-war.zip \
-  --build-arg STABLE_PLUGIN_URL=https://build.geoserver.org/geoserver/2.25.x/ext-latest/ \
-  --build-arg COMMUNITY_PLUGIN_URL=https://build.geoserver.org/geoserver/2.25.x/community-latest/ \
-  --build-arg GS_VERSION=2.24-SNAPSHOT \
-  -t 2.25.x .
+  --build-arg WAR_ZIP_URL=https://build.geoserver.org/geoserver/2.27.x/geoserver-2.27.x-latest-war.zip \
```
buehner, line 71: These examples do not (yet) work as there is not (yet) a `2.27.x` folder: See https://build.geoserver.org/geoserver

README.md
```diff
@@ -226,25 +240,25 @@ Following is the list of the all the environment variables that can be passed do
 | CATALINA_OPTS | Catalina options. Check [ref](https://www.baeldung.com/tomcat-catalina_opts-vs-java_opts) | `-Djava.awt.headless=true` |
 | GEOSERVER_DATA_DIR | Geoserver data directory location | `/opt/geoserver_data/` |
 | GEOSERVER_REQUIRE_FILE | Geoserver configuration used interally | `/opt/geoserver_data/global.xml` |
-| INSTALL_EXTENSIONS | Indicates whether additional GeoServer extensions should be installed | `false` |
+| DOWNLOAD_EXTENSIONS | Indicates whether additional GeoServer extensions should be installed | `false` |
+| INSTALL_EXTENSIONS | Indicates whether additional GeoServer extensions should be installed | `true` |
```
buehner, line 244: The description of `DOWNLOAD_EXTENSIONS` is exactly the same as for `INSTALL_EXTENSIONS`. I guess this was not adjusted after copy and pasting? ;-)

install-extensions.sh
```diff
@@ -4,51 +4,100 @@
...
-    download_extension ${URL} ${EXTENSION}
+    download_extension ${URL} ${EXTENSION} ${GEOSERVER_VERSION}
   done
+  if [ ${#COMMUNITY_EXTENSIONS} -gt 0 ]; then
+    # build community version string from GEOSERVER_VERSION by removing the last part and adding SNAPSHOT
+    if [ -z "$COMMUNITY_EXTENSIONS_VERSION" ]; then
+      COMMUNITY_EXTENSIONS_VERSION="${GEOSERVER_VERSION/-SNAPSHOT/.x}"
+      COMMUNITY_EXTENSIONS_VERSION="${COMMUNITY_EXTENSIONS_VERSION%.*}-SNAPSHOT"
+    fi
+    COMMUNITY_PLUGIN_BASE_URL=${COMMUNITY_PLUGIN_BASE_URL:-"https://build.geoserver.org/geoserver"}
```
buehner, line 55: Unlike others (like `COMMUNITY_PLUGIN_URL`) this new env `COMMUNITY_PLUGIN_BASE_URL` is not declared in the Dockerfile. I think it makes sense to also add it there for the sake of completeness (and to avoid confusion)

install-extensions.sh
```diff
@@ -4,51 +4,100 @@
...
+    if [ -z "$GEOSERVER_RELEASE_BRANCH" ]; then
+      GEOSERVER_RELEASE_BRANCH="${GEOSERVER_VERSION/-SNAPSHOT/.x}"
+      GEOSERVER_RELEASE_BRANCH="${GEOSERVER_RELEASE_BRANCH%.*}.x"
+
+    fi
+    COMMUNITY_PLUGIN_URL=${COMMUNITY_PLUGIN_URL:-"${COMMUNITY_PLUGIN_BASE_URL}/${GEOSERVER_RELEASE_BRANCH}/community-latest/"}
```
buehner, line 61: In general: wouldn't it make sense to set the meaningful/working defaults for `COMMUNITY_PLUGIN_BASE_URL` and `COMMUNITY_PLUGIN_URL`  in the Dockerfile, so we would not have to set them here?

buehner (commented):
I finally found some time to have a look at this and added some first comments. Will try to get this running in a next step. But in general I really like these enhancements.

Could you already rebase your branch against master? There are conflicts currently...

buehner:
> I would be very unhappy with installing community modules into a release intended for production

Maybe this is a misunderstanding? I don't think there is any intention to build community plugins into images. One intention of this PR is to optionally allow users to include experimental community plugins on startup/boot time of the container

aaime:
> I would be very unhappy with installing community modules into a release intended for production - the point of them is to get feedback on snapshot / nightly build releases to support development. 

This position is your personal ideology Jody, but has nothing to do with the GeoServer project at a large. Since community module existed, users have been free to install community modules in releases as they pleased, in their bin, win installers and war packages.

In the docker image as it stands, an artificial limitation has been added. My suggestion is simple: stop treating users as babies. Mixing release and community modules can result in a GeoServer not starting any longer, it is possible and it has actually happened... but as long as the users are made aware of the risks (stick it in bold font in the docs) it's fine. No need to be patronizing.

jodygarnett:
@mbheinen I think what we need is a proposal as this is an on going point of friction.

I also feel a bit bad because I am reporting the policy as I understand it; and am not against changing to something better.

Rather than talk - here is a proposal for discussion: https://github.com/geoserver/geoserver/wiki/GSIP-233

jodygarnett:
Would you like to have a go at fixing the script?

We are also starting to collect "pending" community modules to release alongside GeoServer.

Rikuoja:
@jodygarnett Checked this PR in more detail, it seems like the community plugin url in this PR should work fine, though I haven't tested the script myself. So maybe just consider this a +1 from a user that this PR should fix installation issues bothering me at the moment.

PR #109, opened 2024-10-10
The default cors.allowed.headers `*` is not working 

boeckMt:
On my Setup `Rocky Linux v9.1` with [Podman](https://podman.io/) the default cors.allowed.headers `*` was not working.
The `*` for this is also not in the [tomcat-9.0-doc](https://tomcat.apache.org/tomcat-9.0-doc/config/filter.html#CORS_Filter) and I was not able to find it in [CorsFilter.java](https://github.com/apache/tomcat/blob/9.0.x/java/org/apache/catalina/filters/CorsFilter.java#L656) so I set it to [cors.allowed.headers Defaults](https://tomcat.apache.org/tomcat-9.0-doc/config/filter.html#CORS_Filter) from the documentation.


buehner:
Thank you very much. Makes sense to me. I will merge.

PR #111, opened 2024-10-14
update readme with information about credential overwrite issue

petermorstn:
There is a bug after the restart of the docker container. If the env variable `GEOSERVER_ADMIN_USER` and `GEOSERVER_ADMIN_PASSWORD` are set, new added users and roles are overwritten after a restart. You can avoid this, if you unset the vars after the first run.
I don't know if this is intentional, but perhaps it could be fixed on the code side?


buehner:
This was introduced in https://github.com/geoserver/docker/pull/57

Are you aware of this behaviour @dmenneck ? And do you see a smart way to improve/secure the behavior here?

dmennecke1:
@buehner That certainly sounds like a dangerous feature. One way to fix the problem would be to automatically remove the environment variables at the end of the startup script:

```sh
unset GEOSERVER_ADMIN_PASSWORD
unset GEOSERVER_ADMIN_USER
```

PR #120, opened 2024-11-12
feat: allow EXTRA_JAVA_OPTS to be configured at runtime

alir:
`EXTRA_JAVA_OPTS` is currently non-configurable and is added to `CATALINA_OPTS` at build time.

Users may want to configure the JVM heap values as well as some [application properties](https://docs.geoserver.org/main/en/user/configuration/properties/index.html#application-properties) at runtime, outside of the build process.

This change allows commands like the following to work:

```bash
docker run -p 8080:8080 -e "EXTRA_JAVA_OPTS=-Xmx2g" geoserver
```

This is picked up by Catalina when starting

image

and is also reflected in the GeoServer web console

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,7 +60,8 @@ ENV WEBAPP_CONTEXT=geoserver
 #    --add-opens=java.desktop/sun.java2d.pipe=ALL-UNNAMED
 
 # see https://docs.geoserver.org/stable/en/user/production/container.html
-ENV CATALINA_OPTS="\$EXTRA_JAVA_OPTS \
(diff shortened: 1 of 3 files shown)


buehner:
Hi @alir . Thanks for the PR, but I do not understand the value your change adds as this feature is already available and working. Just try

`docker run -it -p 80:8080 -e "EXTRA_JAVA_OPTS=-Xmx2g" docker.osgeo.org/geoserver:2.26.1` (so your change is not included) and it works out of the box.

In my eyes, there is no need to shift this to the startup script. In case you have a scenario where this is not working, please describe it and we can this up again.

PR #121, opened 2024-11-19
fixed community extension URL

perdiesman:
This updates the community extension download URL as the community extensions have a different naming convention from the stable extensions.


buehner:
I think this already adressed in https://github.com/geoserver/docker/pull/102

perdiesman:
Looks like it is, sorry I must have missed that PR somehow

PR #135, opened 2025-01-20
GEOS-11684: base image does not include GDAL

stacy-rendall:
Further notes/discussion: https://osgeo-org.atlassian.net/browse/GEOS-11684

Since the base image was changed from Ubuntu to Tomcat it no longer includes GDAL, which is required for GDAL/OGR extensions.

The rationale behind this PR is to enable installing GDAL without affecting (i.e. increasing the size of) the initial image.

This PR:
 - adds/documents an `INSTALL_GDAL` environment variable 
 - prior to installing extensions in `startup.sh`:
     - check if INSTALL_GDAL=true
     - check if `gdal-bin` is already installed
     - modify permissions on `/tmp` to enable apt-get update (initially the directory is not writable, which will cause update to fail)
     - update and install gdal-bin, including recommended packages (notably proj)
     - restore `/tmp` permissions

Happy to discuss if an alternative approach might be more appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ The ``startup.sh`` script allows some customization on startup:
 
 * ``INSTALL_EXTENSIONS`` to ``true`` to download and install extensions
 * ``STABLE_EXTENSIONS`` list of extensions to download and install
+* ``INSTALL_GDAL`` to ``true`` to download and install GDAL (required for GDAL/OGR extensions)
 * ``CORS_ENABLED`` to ``true`` to enable CORS support. The following environment variables can be used to customize the CORS configuration.
   * ``CORS_ALLOWED_ORIGINS`` (default ``*``)
   * ``CORS_ALLOWED_METHODS`` (default ``GET,POST,PUT,DELETE,HEAD,OPTIONS``)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -100,6 +100,7 @@ The ``startup.sh`` script allows some customization on startup:
 
 * ``INSTALL_EXTENSIONS`` to ``true`` to download and install extensions
 * ``STABLE_EXTENSIONS`` list of extensions to download and install
+* ``INSTALL_GDAL`` to ``true`` to download and install GDAL (required for GDAL/OGR extensions)
```
aaime, line 103: Would it enable usage of the GDAL extensions though? AFAIK the gdal-java JNI binding library is needed to support that case, which is no longer packaged in Debian based distributions, see this [OSGeo-Live discussion](https://trac.osgeo.org/osgeolive/ticket/2288) for links and workarounds.

buehner: I'd also like to know whether `apt-get install -y gdal-bin` already solves your problem and makes your extension (OGR based WFS Output Format) fully work. Can you confirm this @stacy-rendall ?

I never really had such a use case myself, but I know from others - like @aaime also mentioned - that the GDAL Java bindings are required for some GDAL based extensions/scenarios, which requires some more work as described here:

* https://gdal.org/en/stable/api/java/index.html
* https://gdal.org/en/stable/development/building_from_source.html#building-from-source-java

This could also blow up a resulting image. It would be great to find a solution with full GDAL support. Maybe it also makes sense to provide two GS Docker Image flavors then:

* the basic one (like now)
* an GS Docker Image with full GDAL support (incl. bindings) (-> bigger footprint)

I will try to find some time soon to dive deeper into this GDAL world...

stacy-rendall: Just tested it all again with a clean image: `apt-get install -y gdal-bin` alone is enough for the [ogr-wfs](https://docs.geoserver.org/stable/en/user/extensions/ogr.html) extension to _basically_ work - it only seems to require that "ogr2ogr is available in the path".

To be clear: it is working equivalently for my use case as what it was with the image <= 2.25.3, i.e. prior to the base image changing.

Not sure if I can claim that it is "fully working" - I can't seem to get the extension to transform SRS on any version I have tested (2.25.3 or 2.25.5) - but I'm not sure if that is due to me or an issue with the extension.

buehner:
I found some time for the GDAL Java Bindings. See here: https://github.com/geoserver/docker/pull/138

buehner:
I will close this PR in favor of https://github.com/geoserver/docker/pull/138

The branch has been merged, and the build has been adjusted, so the first `-gdal` version is now available:

```
docker run --env INSTALL_EXTENSIONS=true --env STABLE_EXTENSIONS=gdal -p 8080:8080 docker.osgeo.org/geoserver:2.27.x-gdal
```

From now on, all future releases will include a variant with the `-gdal` suffix.

PR #138, opened 2025-01-23
Add GDAL with Java Bindings Support

buehner:
This pull request allows the integration of **GDAL** into the resulting GeoServer docker image. With a specific focus on **building the Java bindings**. Unlike the default Ubuntu package sources (like `gdal-bin`), which do not only provide an outdated version, but especially do not provide the Java bindings, this implementation ensures full compatibility with GDAL.

Related links:

* https://osgeo-org.atlassian.net/browse/GEOS-11684
* https://github.com/geoserver/docker/pull/135
* https://trac.osgeo.org/osgeolive/ticket/2288
* https://gdal.org/en/stable/api/java/index.html
* https://github.com/OSGeo/gdal/tree/master/docker/ubuntu-small (credits go here for the gdal build stage)

Usage: `docker build --build-arg BUILD_GDAL=true -t my-geoserver-image .`

Note: On average hardware, this takes ~10-15 mins, which is the reason to set the default for `BUILD_GDAL` to `false` (which results in the same docker image as before). There are two more new build arguments for optional use: `PROJ_VERSION=9.5.1` and `GDAL_VERSION=3.10.1`

The use of  `--build-arg BUILD_GDAL=true` leads to an increase of the resulting image of (only!) ~200MB as the installation of gdal requires some dependent packages to be installed, but the footprint is still "minimal" as we pick the build artifacts with `COPY --from=gdal_builder`.

I also updated the release script to build two variants of the image (one without and one with gdal (which has a `-gdal` suffix)).

## Comparison "Before vs After"

Use this to start your image with the gdal extension, which can be used to test the gdal functionality:
`docker run -it -p 8080:8080 --env INSTALL_EXTENSIONS=true --env STABLE_EXTENSIONS="gdal" my-geoserver-image`

### Before (without GDAL Java Bindings)

GeoServer-Log:
```
Native library load failed.
java.lang.UnsatisfiedLinkError: no gdalalljni in java.library.path: /usr/local/lib:/usr/local/tomcat/native-jni-lib:/usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
23-Jan-2025 11:40:07.783 WARNING [main] it.geosolutions.imageio.gdalframework.GDALUtilities.loadGDAL Failed to load the GDAL native libs. This is not a problem unless you need to use the GDAL plugins: they won't be enabled.
java.lang.UnsatisfiedLinkError: 'void org.gdal.gdal.gdalJNI.AllRegister()'
```

no-bindings

### After (with GDAL Java Bindings)

GeoServer-Log:
```
23-Jan-2025 11:38:24.013 INFO [main] it.geosolutions.imageio.gdalframework.GDALUtilities.loadGDAL GDAL Native Library loaded (version: 3.10.1)
```

bindings

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,178 @@
-FROM tomcat:9.0.102-jdk17-temurin-jammy@sha256:7ecf106b1c45baf845856e11c9c2ff7c65d54a9b015ef7e86288293c197d0162
+ARG BUILDER_BASE_IMAGE=eclipse-temurin:17-jdk-jammy
+ARG GEOSERVER_BASE_IMAGE=tomcat:9.0.102-jdk17-temurin-jammy@sha256:7ecf106b1c45baf845856e11c9c2ff7c65d54a9b015ef7e86288293c197d0162
+
+ARG GS_VERSION=2.26.2
+ARG BUILD_GDAL=false
+ARG PROJ_VERSION=9.5.1
+ARG GDAL_VERSION=3.10.2
+ARG INSTALL_PREFIX=/usr/local
+
+# This is a multi stage build.
+
+# The gdal_builder image/stage can be enabled by using
+# `--build-arg BUILD_GDAL=true`. The gdal_builder is derived
+# from the osgeo/gdal Dockerfile (small ubuntu) and simplified
+# version activating the JAVA features on top.
+# It builds gdal from sources, which is necessary to get
+# working gdal JAVA Bindings, e.g. for the geoserver gdal extension.
+# see https://trac.osgeo.org/osgeolive/ticket/2288
+#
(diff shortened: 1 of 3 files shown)


huoluzi:
we need it!

hwbllmnn (approved):

buehner:
> we need it!

@huoluzi The branch has been merged, and the build has been adjusted, so the first `-gdal` version is now available:

```
docker run --env INSTALL_EXTENSIONS=true --env STABLE_EXTENSIONS=gdal -p 8080:8080 docker.osgeo.org/geoserver:2.27.x-gdal
```

From now on, all future releases will include a variant with the `-gdal` suffix.

PR #203, opened 2025-09-20
Update base images to noble

hwbllmnn:
This updates the base images to Ubuntu 24 LTS.


buehner:
I added two (non-changing) suggestions to have the images more explicit (renovate will care about future updates).

I could successfully build and run the basic and the gdal variant.

Feel free to merge after applying the suggestions

buehner (changes requested):
One more thought: We should probably switch from `apt-get` to `apt` everywhere

PR #306, opened 2026-07-24
fix: extension prefix collison

buehner:
a potential fix for https://osgeo-org.atlassian.net/browse/GEOS-12131


solita-jhaakana:
Thanks for working on this! I encountered what appears to be this exact issue with the `3.1.x` image: installing both `wps` and `geofence-wps` caused `geofence-wps` to be extracted twice, leaving the core WPS classes missing. Renaming the WPS archive to avoid the prefix collision resolved it. If this PR covers that case, it would be great to see it merged.

